PR #101, opened 2022-03-08
Add translation for local table

rtpsw:


ibis_substrait/compiler/translate.py
```diff
@@ -583,6 +583,36 @@ def unbound_table(
     )
 
 
+@translate.register(ops.LocalTable)
```
cpcloud, line 586: `ibis-substrait` currently depends on ibis 2.x, so this is a non-starter as is.

I would suggest implementing `LocalTable` in `ibis-subtrait` and not in `ibis`. We can always push it back into `ibis` upstream once the implementation is fleshed out.

ibis_substrait/compiler/translate.py
```diff
@@ -583,6 +583,36 @@ def unbound_table(
...
+def local_table(
+    op: ops.LocalTable, expr: ir.TableExpr, _: SubstraitCompiler, **kwargs: Any
+) -> stalg.Rel:
+    def path_type_of(name : str | None):
```
cpcloud, line 590: These functions should be defined outside of the `local_table` function with `_` prefixed to the name (`_path_type_of` instead of `path_type_of`) to indicate that they are not public APIs.

ibis_substrait/compiler/translate.py
```diff
@@ -583,6 +583,36 @@ def unbound_table(
...
+        else:
+            return "uri_path_glob" if name.contains("*") else "uri_path"
+    def format_of(name : str | None):
+        return (stalg.ReadRel.LocalFiles.FileOrFiles.FileFormat.FILE_FORMAT_PARQUET
+                if name.endswith(".parquet")
+                else stalg.ReadRel.LocalFiles.FileOrFiles.FileFormat.FILE_FORMAT_UNSPECIFIED)
+    return stalg.Rel(
+        read=stalg.ReadRel(
+            # TODO: filter,
+            # TODO: projection,
```
cpcloud, line 604: If you can make an issue for these so we can discuss, that would be great. Ideally we don't have any `TOTO`s without associated issues. Otherwise, if you don't feel like creating tickets please remove the `TODO` and add a small description to the comment.

ibis_substrait/compiler/translate.py
```diff
@@ -583,6 +583,36 @@ def unbound_table(
...
+def local_table(
+    op: ops.LocalTable, expr: ir.TableExpr, _: SubstraitCompiler, **kwargs: Any
+) -> stalg.Rel:
+    def path_type_of(name : str | None):
+        if name is None:
+            raise ValueError("local table name is None")
+        if name.startswith("file://"):
+            return "uri_folder" if name.endswith("/") else "uri_file"
+        else:
+            return "uri_path_glob" if name.contains("*") else "uri_path"
```
cpcloud, line 596:
```suggestion
            return "uri_path_glob" if glob.has_magic(name) else "uri_path"
```
You'll need to `import glob` at the top of the file as well.

ibis_substrait/compiler/translate.py
cpcloud, line 590:
```suggestion
    def path_type_of(name : str | None) -> str:
```

ibis_substrait/compiler/translate.py
```diff
@@ -583,6 +583,36 @@ def unbound_table(
...
 
+@translate.register(ops.LocalTable)
+def local_table(
+    op: ops.LocalTable, expr: ir.TableExpr, _: SubstraitCompiler, **kwargs: Any
+) -> stalg.Rel:
+    def path_type_of(name : str | None):
+        if name is None:
+            raise ValueError("local table name is None")
+        if name.startswith("file://"):
+            return "uri_folder" if name.endswith("/") else "uri_file"
```
cpcloud, line 594: What should happen when `name` is a file but a user has specified a trailing `/`, and similarly when a `name` is a directory but a user hasn't specified a trailing `/`?

cpcloud (changes requested):
Thanks! Few comments to address.

ibis_substrait/compiler/translate.py
rtpsw: Actually, I [copied](https://github.com/ibis-project/ibis-substrait/blob/94980bcbe5932666ec50698466be369ed5859117/ibis_substrait/compiler/translate.py#L578-L579) these. Presumably there are issues for them that can be referenced?

ibis_substrait/compiler/translate.py
rtpsw: I wasn't sure about this myself. IIUC, this point in the code is just a declaration, and the file (or directory) is not being opened at this time and may not even exist yet. So, there should be some way for the caller to signal a file vs a directory. My current proposal has this signal embedded within the name. A possible alternative is to add a parameter to the `local_table` operation.

ibis_substrait/compiler/translate.py
rtpsw: OK. Should I cancel this PR and open a new one in Ibis? Or should we flush out the remaining issues here first?

cpcloud: Let's finish up the existing comments and then move it out.

ibis_substrait/compiler/translate.py
cpcloud: Ah, ok. It's fine to leave them for now

rtpsw:
Following some offline discussions, it's not clear to me whether this is the API desired long-term, even if all related code moves to ibis-substrait, so I'll cancel this. If you nevertheless would like this API to move forward, please let me know.

PR #125, opened 2022-03-24
chore: add protobuf dependency

gforsyth:
discovered that you can't use the compiler unless we've also installed `protobuf` (which, you know, makes sense)


cpcloud:
Thank you!

jreback:
hmm nbd but this now a breaking change

cpcloud:
What breaks?

jreback:
adding a new dependency is by definition a breaking change

PR #177, opened 2022-04-27
fix: dispatch underlying op for aliases

gforsyth:
`ops.Alias` was added in Ibis recently and the dispatcher was defaulting
to `ops.ValueOp` when it received one -- instead this added in a
specific Alias dispatch rule to pull out the underlying op.

xref ibis-project/ibis#3276

### DIFF
--- a/ibis_substrait/compiler/decompile.py
+++ b/ibis_substrait/compiler/decompile.py
@@ -26,6 +26,7 @@ from ..proto.substrait import algebra_pb2 as stalg
 from ..proto.substrait import plan_pb2 as stp
 from ..proto.substrait import type_pb2 as stt
 from .core import SubstraitDecompiler, _get_fields, which_one_of
+from .mapping import SUBSTRAIT_IBIS_OP_MAPPING
 from .translate import _MICROSECONDS_PER_SECOND, _MINUTES_PER_HOUR, _SECONDS_PER_MINUTE
 
 T = TypeVar("T")
@@ -730,7 +731,7 @@ def _decompile_expression_scalar_function(
 ) -> ir.ValueExpr:
     extension = decompiler.function_extensions[msg.function_reference]
     function_name = extension.name
-    op_type = getattr(ops, inflection.camelize(function_name))
+    op_type = getattr(ops, SUBSTRAIT_IBIS_OP_MAPPING[function_name])
     args = [decompile(arg, children, field_offsets, decompiler) for arg in msg.args]
     expr = op_type(*args).to_expr()
     output_type = _decompile_type(msg.output_type)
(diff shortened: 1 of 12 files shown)


ibis_substrait/compiler/translate.py
```diff
@@ -441,6 +441,17 @@ def _translate_window_bounds(
     return translate_preceding(*preceding), translate_following(*following)
 
 
+@translate.register(ops.Alias)
```
cpcloud, line 444: Should we do this in a backwards compatible way? E.g., importing `Alias` and only registering if the import succeeds?

gforsyth: hmm, it's a good question.  I'm not sure what, if any, backwards compatibility guarantees we're prepared to make at this point.  We could make it backwards compatible (at least for now) just to demonstrate that we can, I suppose.  
I realize I'm not even certain if `ibis-substrait` is compatible with ibis 2.1.1

gforsyth: Roger that.  I'll add in some version gating

gforsyth:
@cpcloud -- are the "expected checks" here not getting picked up because I changed the build matrix?  Not sure what they are waiting on

cpcloud (approved):
@gforsyth Yep, that's right. This LGTM. I'll merge this after #185.

cpcloud:
Will adjust the required checks after merging this.

pyproject.toml
```diff
@@ -20,9 +20,11 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
-ibis-framework = "^2.0.0"
+ibis-framework = "3.0.2"
```
cpcloud, line 23: Does it make sense to make this `>=2,<4`?

ibis_substrait/compiler/decompile.py
```diff
@@ -730,7 +731,7 @@ def _decompile_expression_scalar_function(
 ) -> ir.ValueExpr:
     extension = decompiler.function_extensions[msg.function_reference]
     function_name = extension.name
-    op_type = getattr(ops, inflection.camelize(function_name))
```
cpcloud, line 733: Can we remove the `inflection` dependency in this PR?

pyproject.toml
gforsyth: yes -- I had this hard-pinned because of the now-revoked versions that were misreporting `__version__`

ibis_substrait/compiler/decompile.py
gforsyth: done

Issue #181, opened 2022-04-29
Fix missing timezone parameter in timestamp translation


rtpsw:
The timezone parameter of `TimestampTZ` is [missing here](https://github.com/ibis-project/ibis-substrait/blob/cdb651fcbc993a36503d024e55d3da6cda24a3e1/ibis_substrait/compiler/translate.py#L109).

rtpsw:
Cancelled per the discussion in https://github.com/substrait-io/substrait/issues/190 .

PR #182, opened 2022-04-29
feat(proto): fix missing timezone parameter in timestamp translation

rtpsw:
See https://github.com/ibis-project/ibis-substrait/issues/181


rtpsw:
@cpcloud, it looks like `commitlint` fails due to checking more than just the last commit message, which I thought would be the one to be selected in the eventual squashing. How to handle this error?

cpcloud:
This looks like a custom extension to substrait and a modification of the spec outside of the main substrait-io/substrait repo.

Regarding the actual change, the `TimetampTZ` type is intentionally without a timezone. It only indicates that there is a timezone. The design is similar to the way time zones work in most relational databases like postgres (see https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-TIMEZONES).

That said, unlike postgres there's nothing documented in substrait saying that a `TimestampTZ` is assumed to be UTC so we should nail that down.

rtpsw:
Thanks, I wasn't aware that spec changes need to go to substrait-io/substrait. 

Given `TimezoneTZ` does not carry a timezone, how should a timezone value be represented within a Substrait plan? I think such a timezone value is needed for casting of a time-string to a specific timezone. Currently, in [related Arrow code](https://github.com/rtpsw/arrow/blob/f8d6eac3d97876eed43faa9aaec9f1201dec55ba/cpp/src/arrow/engine/substrait/expression_internal.cc#L224-L226), I see a default timezone value is used.

I'm OK with canceling this PR if the conclusion is that a timezone value is not needed and only the Substrait docs need to be fixed.

rtpsw:
@cpcloud - ping. See question [here](https://github.com/ibis-project/ibis-substrait/pull/182#issuecomment-1115033129).

cpcloud:
@rtpsw Is there a use case you have that requires a change to the substrait spec? If so, let's discuss in the substrait-io repo and close this PR out! Thanks!

rtpsw:
Created https://github.com/substrait-io/substrait/issues/190 for the discussion.

Issue #184, opened 2022-05-01
Expand table expression to its columns


rtpsw:
This is a proposal for making the handling of Substrait selections easier for Substrait consumers. Currently, Substrait consumers, such as Arrow, need to handle schema computations during Substrait deserialization in order to infer the columns of a table expression, which are needed for setting up a projection. In Arrow, such computations are not supported because schemata instances are only created after deserialization. By having Ibis-Substrait expand the table expression to its columns, a Substrait consumer can directly construct the corresponding projection.

cpcloud:
Closed by #185.

PR #185, opened 2022-05-01
feat(compiler): expand table expression to its columns

rtpsw:
See https://github.com/ibis-project/ibis-substrait/issues/184

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -647,6 +647,13 @@ def selection(
         )
 
     # projection
+    selections = [
+        col
+        for sel in op.selections
+        for col in (
+            sel.get_columns(sel.columns) if isinstance(sel, ir.TableExpr) else [sel]
+        )
+    ]
     if op.selections:
         relation = stalg.Rel(
             project=stalg.ProjectRel(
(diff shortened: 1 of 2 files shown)


rtpsw:
Note that one alternative is to make the table expansion a choice controlled by a parameter of `ibis.mutate` placed in a member of `ops.Selection`.

ibis_substrait/compiler/translate.py
```diff
@@ -647,6 +647,11 @@ def selection(
         )
 
     # projection
+    selections = [
+        x
+        for s in op.selections
+        for x in (s.get_columns(s.columns) if isinstance(s, ir.TableExpr) else [s])
```
cpcloud, line 653:
```suggestion
        col
        for sel in op.selections
        for col in (sel.get_columns(sel.columns) if isinstance(sel, ir.TableExpr) else [sel])
```

cpcloud (changes requested):
@rtpsw Can you add a test for this change?

rtpsw:
> @rtpsw Can you add a test for this change?

Will do.

cpcloud:
@rtpsw This looks reasonable to me.

cpcloud (approved):
LGTM, just needs a squash then it's good to go.

PR #215, opened 2022-05-13
fix: allow for ops.Contains with only one value

gforsyth:


ibis_substrait/compiler/translate.py
```diff
@@ -937,9 +937,13 @@ def _contains(
     compiler: SubstraitCompiler,
     **kwargs: Any,
 ) -> stalg.Expression:
+    if not isinstance(op.options, (collections.abc.Sequence, ir.ListExpr)):
```
cpcloud, line 940: Is this to support column expressions? If so, I'm not sure if `SingularOrList` is the right mapping. I think you'd probably want a `Subquery` since the corresponding SQL would be something like

```sql
SELECT a IN (SELECT b FROM t)
FROM t
```

That would be something like:

```python
Subquery(
    in_predicate=InPredicate(
        needles=[translate(op.value, compiler, **kwargs)],
        haystack=translate(op.options.to_projection(), compiler, **kwargs),
    )
)
```

cpcloud (changes requested):

ibis_substrait/compiler/translate.py
gforsyth: It's coming up in situations like this one:
```
In [16]: ~partsupp.ps_suppkey.isin(
    ...:     supplier.filter([supplier.s_comment.like("%Customer%Complaints%")]).s_suppkey
    ...: )
Out[16]: 
r0 := UnboundTable: supplier
  s_suppkey   int32
  s_name      string
  s_address   string
  s_nationkey int32
  s_phone     string
  s_acctbal   decimal(prec=15, scale=2)
  s_comment   string

r1 := UnboundTable: partsupp
  ps_partkey    int32
  ps_suppkey    int32
  ps_availqty   int32
  ps_supplycost decimal(prec=15, scale=2)
  ps_comment    string

r2 := Selection[r0]
  predicates:
    r0.s_comment LIKE '%Customer%Complaints%'

Not(Contains(value=r1.ps_suppkey, options=r2.s_suppkey))
```

so there's only a single "option" but that option is a selection

gforsyth: and looking at your example again, I see that yes, we probably want a subquery. 🥴

ibis_substrait/compiler/translate.py
```diff
@@ -937,12 +937,29 @@ def _contains(
...
+        # Column expression
+        return stalg.Expression(
+            subquery=stalg.Expression.Subquery(
+                in_predicate=stalg.Expression.Subquery.InPredicate(
+                    needles=[translate(op.value, compiler, **kwargs)],
+                    haystack=translate(
+                        op.options.to_projection(),
+                        compiler,
+                        # if we to a `to_projection` do we need to mess
+                        # with field offsets?
```
cpcloud, line 950: I don't think so, but we'll have to handle outer references for correlated subqueries at some point. We can do that as a concerted effort in a follow-up PR.

gforsyth: Ok -- does it make sense that we wouldn't pass through the `**kwargs` here?  Because if we do, 3 of the TPC queries fail because they are passing non-zero child offsets to `ops.Selection`

```
AssertionError: non-empty child_rel_field_offsets passed in to selection translation rule
```

where the `child_rel_field_offsets` in the `kwargs` are:
```
(Pdb) p child_rel_field_offsets
{UnboundTable(schema=ibis.Schema {
  ps_partkey     int32
  ps_suppkey     int32
  ps_availqty    int32
  ps_supplycost  decimal(prec=15, scale=2)
  ps_comment     string
}, name='partsupp'): 0, UnboundTable(schema=ibis.Schema {
  p_partkey      int32
  p_name         string
  p_mfgr         string
  p_brand        string
  p_type         string
  p_size         int32
  p_container    string
  p_retailprice  decimal(prec=15, scale=2)
  p_comment      string
}, name='part'): 5}
```

PR #231, opened 2022-05-23
feat: add ops.ExistsSubquery

gforsyth:


gforsyth:
This works locally -- I haven't quite got the decompilation working and it also relies on ibis-master, so this can sit as a draft for a bit

PR #252, opened 2022-06-27
feat: add support for ops.Extract<span>

gforsyth:
Covers `ExtractYear`, `ExtractMonth`, and `ExtractDay`.  Substrait also
allows `SECONDS` as the specified argument but I'm not clear on exactly
what they're looking for there.  Epoch Seconds?


cpcloud:
Usually there's a `SECONDS` or similar for extracting the seconds component of a timestamp or time, and an `EPOCH_SECONDS` or similar for converting timestamps to seconds since the UNIX epoch.

gforsyth:
> Usually there's a `SECONDS` or similar for extracting the seconds component of a timestamp or time, and an `EPOCH_SECONDS` or similar for converting timestamps to seconds since the UNIX epoch.

Yeah, I guess I'm confused about the breakdown being Y-M-D and seconds (without hours / minutes a la ISO 8601).

PR #253, opened 2022-06-28
fix(extract): substrait function argument ordering

gforsyth:
According to the spec, you cannot have the function option argument
specified after a "regular" argument.

Definitely time to add the validator to our CI runs

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -968,16 +968,18 @@ def _extractdatefield(
     compiler: SubstraitCompiler,
     **kwargs: Any,
 ) -> stalg.Expression:
+    # e.g. "ExtractYear" -> "YEAR"
+    span = type(op).__name__[len("Extract") :].upper()
+    args = (
+        translate(arg, compiler, **kwargs)
+        for arg in op.args
+        if isinstance(arg, ir.Expr)
+    )
+
     scalar_func = stalg.Expression.ScalarFunction(
         function_reference=compiler.function_id(expr),
         output_type=translate(expr.type()),
-        args=[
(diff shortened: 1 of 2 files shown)


ibis_substrait/compiler/translate.py
```diff
@@ -968,16 +968,19 @@ def _extractdatefield(
     compiler: SubstraitCompiler,
     **kwargs: Any,
 ) -> stalg.Expression:
+    # e.g. "ExtractYear" -> "YEAR"
+    span = type(op).__name__.lstrip("Extract").upper()
```
cpcloud, line 972: I've always found `{l,r}strip` to have surprising behavior:

```
>>> 'ExtractYear'.lstrip(''.join(reversed('tcartxE')))
'Year'
>>> 'ExtractYear'.lstrip('tcartxE')
'Year'
```

Another way might be `type(op).__name__[len("Extract"):]`

Not a big deal though.

ibis_substrait/compiler/translate.py
```diff
@@ -968,16 +968,19 @@ def _extractdatefield(
...
-            translate(arg, compiler, **kwargs)
-            for arg in op.args
-            if isinstance(arg, ir.Expr)
-        ],
     )
-    # e.g. "ExtractYear" -> "YEAR"
-    span = type(op).__name__.lstrip("Extract").upper()
     scalar_func.args.add(enum=stalg.Expression.Enum(specified=span))
+    for arg in args:
+        scalar_func.args.append(arg)
```
cpcloud, line 985:
```suggestion
    scalar_func.args.extend(args)
```

cpcloud (approved):

PR #256, opened 2022-06-30
fix: translate all literals as nullable

gforsyth:
We need to specify nullability in all literals -- I'm not sure if just defaulting to "YES" is the right move here, but that's definitely my starting point.


cpcloud:
I guess logically `False` feels like the correct thing since a non-null literal can't be anything but what it is and on the other hand `Literal(None, type=int8)` would be nullable. I'm pretty sure that in ibis if we were to change this (we currently use `nullable=True` for literal value types) that it would break everything because we don't have unification of null and non-null types completely sorted out.

I'm not sure what the downsides are to defaulting to nullable literals are but I think we can start with this.

cpcloud (approved):

PR #258, opened 2022-07-01
Add TPC-H query compilation + validation tests

gforsyth:
Adds in compilation tests for all 22 TPC-H queries and also runs the substrait-validator against the output.

Currently silencing several warnings in the validator, we can look into those more closely to figure out which, if any, we care deeply about.

This will fail to pass CI until Ibis 3.1 is released.


cpcloud:
Should this be closed in light of #292?

gforsyth:
Yes!

PR #278, opened 2022-07-20
chore(substrait): update substrait to version 0.8

gforsyth:


cpcloud (approved):
Should this be a `fix` commit to bump the version?

gforsyth:
> Should this be a `fix` commit to bump the version?

I didn't realize that the auto-release was predicated on the commit types, but that makes sense.  Rebased and updated

PR #289, opened 2022-07-28
refactor: replace deprecated `args` with `arguments`

gforsyth:
Also add required support for `substrait.algebra_pb2.FunctionArgument`
which is the protobuf message type required for arguments to
`ScalarFunction` and `AggregateFunction`


saulpw (approved):
Looks good to me!

PR #290, opened 2022-07-28
feat: add support for ReadRel.LocalFiles

lidavidm:


lidavidm:
Mostly a mechanical binding of the current Substrait proto. Helper classes were added to reflect the oneofs.

ibis_substrait/expr/operations/relations.py
```diff
@@ -0,0 +1,96 @@
...
+@validator  # type: ignore[misc]
+def valid_uri(arg: str, **kwargs: typing.Dict[typing.Any, typing.Any]) -> str:
+    arg = rlz.instance_of(str, arg)
+    uri = urllib.parse.urlparse(arg)
+    if not uri.scheme:
+        raise ValueError(f"Missing URI scheme in '{arg}'")
+    return arg
+
+
+class FileFormat(Annotable, Comparable):
```
kszucs, line 19: If it is intented to be an ibis operation, then subclass from `ibis.expr.operations.Node`

ibis_substrait/expr/operations/relations.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    Base class representing the format of files in a LocalFiles.
+    """
+
+    def __equals__(self, other: typing.Any) -> bool:
```
kszucs, line 24: If you subclass from `Node` then you can remove this method.

ibis_substrait/expr/operations/relations.py
```diff
@@ -0,0 +1,96 @@
...
+import urllib.parse
+
+import ibis.expr.operations.relations as rel
+import ibis.expr.rules as rlz
+from ibis.common.grounds import Annotable, Comparable
+from ibis.common.validators import validator
+
+
+@validator  # type: ignore[misc]
+def valid_uri(arg: str, **kwargs: typing.Dict[typing.Any, typing.Any]) -> str:
```
kszucs, line 11: Simply call it `uri`?

kszucs:
The translation looks good, but what's the use case of these operations?

lidavidm:
The use case is to be able to represent a query over files local to the execution backend (though, a "file" may very well be something in cloud storage), so that we can support things like Acero/Substrait where there isn't really a table catalog.

We talked offline a bit and I'll rework this to be 1) not Substrait-specific and 2) part of the Ibis core instead of out here.

lidavidm:
I opened https://github.com/ibis-project/ibis/pull/4293

PR #292, opened 2022-07-29
Namespace proto files and add TPC-H tests and validator checks

gforsyth:
Supercedes #258 

This is a bit bigger than I anticipated but I think it makes sense to do it all together, since piecemeal involves several obnoxious workarounds, so:

- Adds TPC-H fixtures and tests
  - First set of tests are strictly "can this even compile?" (not a correctness check)
  -  Second set are "if it can compile, is it valid substrait?" (again, not checking for plan correctness, just validity)
- Adds the substrait validator (for checking plan validity)
  - Note that calling it inline, any warnings cause a failure, so I've changed the level of the warnings to `info` to avoid failure on warnings 
- Namespaces the "vendored" substrait into `proto.substrait.ibis` -- this avoids descriptor pool collisions in-process with (in this case) the validator.  This should also avoid potential collisions in the future with in-process communication.
  - There's some suboptimal scripting (on my part) to handle the shuffle of renaming but not leaving the repo in a dirty state. 
- Updates `google.protobuf` to 3.20.0 so we match what's in the buf spec (and don't collide with substrait-validator)
- Removes Ibis 2.x from the CI runs

I think everything is working, and this should remove the need for setting `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION="python"` which is gross


cpcloud:
@icexelloss Are y'all using ibis v2 with ibis-substrait? If not, we'd like to drop support for it in `ibis-substrait`.

cpcloud (approved):

icexelloss:
@cpcloud If by "ibis v2" you mean version 2.x then no we are not using ibis-substrait with it (currently we are on ibis 3.0 but haven't started to use ibis-substrait internally)

PR #297, opened 2022-08-01
fix(translate): use FunctionArgument in ops.Count translation

gforsyth:
Follow up to #289 -- missed one specification of `FunctionArgument`


gforsyth:
> Is there a low-effort test we can add for this?

Yep!

PR #316, opened 2022-08-10
fix: always calculate child rel field offsets for join predicates

gforsyth:
The confusion for me has been the location of the `field` lookup,
referenced in the
`expression->scalar_function->arguments->value->selection->direct_reference->struct_field`

Where is the list of columns that pokes in to?

It doesn't explicitly exist -- it is the flattened list of all columns
from the
left and the right tables.

We were correctly calculating the `relative_offset` -- the offset within the
left- or right-table -- but our `base_offset` was wrong.  Why?

Because in some situations, we don't have the `child_rel_field_offsets`
to pass to the `translate` call for the _join predicates_ and they need
to be there.

So we check if that has been calculated, and if it hasn't, then we
calculate them and pass them in.

Also, I am _sure_ there are edge cases that this doesn't cover, but this is a distinct improvement.

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -798,6 +798,7 @@ def join(
     compiler: SubstraitCompiler,
     **kwargs: Any,
 ) -> stalg.Rel:
+    child_rel_field_offsets = kwargs.pop("child_rel_field_offsets", None)
     return stalg.Rel(
         join=stalg.JoinRel(
             left=translate(op.left, compiler, **kwargs),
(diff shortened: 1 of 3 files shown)


gforsyth:
Found an edge case -- if we use `named_table.column` as a predicate in a nested join, the field offset is wrong.
Not sure that should block this, but I'll work on fixing that

cpcloud:
> Found an edge case -- if we use `named_table.column` as a predicate in a nested join, the field offset is wrong. Not sure that should block this, but I'll work on fixing that

Can you make a small issue for it?

cpcloud (approved):

PR #323, opened 2022-08-16
fix(joins): compute field offsets for nested joins

### LINKED ISSUE #317: Incorrect field offsets for nested join where predicate points to root table
The following test fails:
```
@pytest.fixture
def r():
    return ibis.table([("a", "string"), ("b", "int64")], name="r")


@pytest.fixture
def q():
    return ibis.table([("e", "string"), ("f", "int64")], name="q")


def test_roundtrip_nested_join(compiler, s, r, q):
    expr = s.join(r, s.c == r.a)
    expr = expr.join(q, r.a == q.e)

    plan = compiler.compile(expr)
    (result,) = decompile(plan)
    assert result.equals(expr)
```

Note that the left side of the predicate in the second join is referencing `r.a` and not `expr.a`

```python
> expr
r0 := UnboundTable: q
  e string
  f int64

r1 := UnboundTable: r
  a string
  b int64

r2 := UnboundTable: s
  c string
  d int64

r3 := InnerJoin[r2, r1] r2.c == r1.a

InnerJoin[r3, r0] r1.a == r0.e
```

```python
> result
r0 := UnboundTable: q
  e string
  f int64

r1 := UnboundTable: r
  a string
  b int64

r2 := UnboundTable: s
  c string
  d int64

r3 := InnerJoin[r2, r1] r2.c == r1.a

InnerJoin[r3, r0] r3.c == r0.e
```

gforsyth:
I _think_ this is the right fix for the failure to handle nested joins.

The decompiled expression isn't identical, but it I believe it is
equivalent.

I think this fixes #317 

The tests are failing because we're comparing the round-tripped expressions and they aren't identical.
Need to think about how to run those checks.

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -671,9 +671,12 @@ def _get_child_relation_field_offsets(table: ir.TableExpr) -> dict[ops.TableNode
     """
     table_op = table.op()
     if isinstance(table_op, ops.Join):
+        # Descend into the left and right tables to grab offsets from nested joins
+        left_keys = _get_child_relation_field_offsets(table_op.left)
+        right_keys = _get_child_relation_field_offsets(table_op.right)
         root_tables = [table_op.left.op(), table_op.right.op()]
         accum = [0, len(root_tables[0].schema)]
-        return dict(zip(root_tables, accum))
+        return {**left_keys, **right_keys, **dict(zip(root_tables, accum))}
     return {}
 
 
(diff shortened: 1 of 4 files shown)


gforsyth:
Ok, I've opted for a slightly gross, but at-this-stage-unavoidable comparison between the SQL representations of the original and the round-tripped expressions.

cpcloud (approved):

cpcloud:
Excellent! Thank you.

cpcloud:
Excellent! Thank you.

PR #338, opened 2022-09-02
fix: remove dev dependencies from core dependencies

### LINKED ISSUE #337: Issue with platformdirs in the dependency
Hi,

I am trying to import ibis-substrait into our build system and hit an issue with `platformdirs`. It seems it is built with "hatchling" and our internal system doesn't understand "hatchling" build system.

After looking at the substrait code, I didn't find where it was being used in the source code, perhaps this is a dev dependency and not a regular dependency?

gforsyth:
Resolves #337


cpcloud (approved):
Sweet, thanks!

PR #366, opened 2022-10-07
feat: add extension URIs

gforsyth:
This is a first cut at properly supporting substrait extensions.

According to https://substrait.io/extensions/#function-signature-compound-names, the function signature should be of the form `<function name>:<short_arg_type0>_<short_arg_type1>_..._<short_arg_typeN>`

Further, we are expected to say which of the extension yaml files to refer to when we create a new `extensionUri`

Towards that, this PR does a few things:
1. vendors in the substrait extension yaml files so we can generate a dictionary of accepted signatures
2. Updates how we generate extension URIs
3. Adds in a lookup step to the `function_id` generation to use the correct extension function signature

TODO:
1. The assumption that substrait makes is that producers will handle any required casting before serialization -- currently if we did something like `1 - lineitem.l_discount` Ibis will infer the type of the literal `1` as an `int8`.  `l_discount` is a decimal, so substrait wants us to cast the `1` to a decimal before serialization.

Also, there's a bunch of gross yaml munging in here that could be a lot nicer, but it works for now.


gforsyth:
superceded by #510

PR #367, opened 2022-10-11
feat: add dispatch for ops.Where

gforsyth:
The latest releases of `ibis` added an `ops.Where` operation that we weren't catching.

Adding dispatch for `ops.Where` here to an `If_Then`


jcrist (approved):
:rocket:

PR #383, opened 2022-10-17
fix(projection): Use emit and not project for column selection

### LINKED ISSUE #365: bug: project rels are not correct
According to the substrait spec, project rels always append, and removal of columns is done through `emit`. Right now, ibis-substrait only keeps the most recently computed expressions around.

We should change the implementation to always keep the columns from the previous relation and, drop them implicitly using `emit`.

gforsyth:
I based this off of the output for similar queries generated by `isthmus`, the `substrait-java` implementation.

I need to figure out how to test this for correctness.

Fixes #365

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -10,6 +10,7 @@ import collections.abc
 import datetime
 import decimal
 import functools
+import itertools
 import operator
 import uuid
 from typing import Any, Mapping, MutableMapping, Sequence, TypeVar
@@ -20,6 +21,7 @@ import ibis.expr.schema as sch
 import ibis.expr.types as ir
 import toolz
 from ibis import util
+from ibis.util import to_op_dag
 
 from ..proto.substrait.ibis import algebra_pb2 as stalg
 from ..proto.substrait.ibis import type_pb2 as stt
@@ -637,6 +639,7 @@ def unbound_table(
         read=stalg.ReadRel(
             # TODO: filter,
             # TODO: projection,
+            common=stalg.RelCommon(direct=stalg.RelCommon.Direct()),
             base_schema=translate(op.schema),
             named_table=stalg.ReadRel.NamedTable(names=[op.name]),
         )
@@ -694,7 +697,6 @@ def selection(
         child_rel_field_offsets=child_rel_field_offsets,
         **kwargs,
     )
-
     # filter
     if op.predicates:
         relation = stalg.Rel(
@@ -709,7 +711,7 @@ def selection(
             )
         )
 
-    # projection
+    # projection / emit
     selections = [
         col
         for sel in op.selections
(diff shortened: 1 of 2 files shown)


ibis_substrait/compiler/translate.py
```diff
@@ -480,6 +482,11 @@ def value_op(
             function_reference=compiler.function_id(expr),
             output_type=translate(expr.type()),
             arguments=[
+                stalg.FunctionArgument(
```
bkietz, line 485: This option doesn't apply to all scalar functions, which I think is what's causing [the bulk of the CI failures](https://github.com/ibis-project/ibis-substrait/actions/runs/3268859832/jobs/5375755332#step:6:1203). I'm not sure what the best ibis::way is to detect what enum args are supported by a function

bkietz (changes requested):

ibis_substrait/compiler/translate.py
gforsyth: doing this "correctly" is going to have to wait until we're properly parsing the extension yaml files to read in the options (or until the upstream PR making optional parameters _optional_ lands).

For now this is only impacting decompilation, so I'm going to xfail them and circle back.

ibis_substrait/compiler/translate.py
```diff
@@ -474,12 +476,25 @@ def value_op(
     compiler: SubstraitCompiler,
     **kwargs: Any,
 ) -> stalg.Expression:
+    error_args = []
+    # TODO(gforsyth): remove this brittle workaround after extension parsing is ready
+    if (
+        isinstance(op, ops.Binary)
+        and not isinstance(op, ops.Comparison)
+        and isinstance(op.left.type(), (dt.Integer, dt.Floating, dt.Decimal))
+        and isinstance(op.right.type(), (dt.Integer, dt.Floating, dt.Decimal))
```
bkietz, line 485:
```suggestion
        isinstance(op, ops.NimericBinary)
```
or [maybe](https://github.com/ibis-project/ibis/blob/4df30e61862d436f5966ca3eb2381e6941d485bc/ibis/expr/operations/numeric.py#L19)
```suggestion
        isinstance(op, (ops.Add, ops.Multiply, *other_error_needing_ops))
```

gforsyth: I went with `NumericBinary` and also added in a check to make sure we add in the required `rounding` argument for non-integers

gforsyth: Scratch that -- `NumericBinary` was added after Ibis 3.0.2

ibis_substrait/compiler/translate.py
gforsyth: @jvanstraten -- that is definitely the better target state.  For the moment, I'm after some small degree of interoperability with pyarrow and setting it to `unspecified` doesn't seem to play nice.

jvanstraten: Huh, based on the current Acero code I thought it should have. Carry on then :)

jcrist (approved):
With my admittedly limited understanding of substrait, this looks like a sufficient fix to me. :+1:

PR #394, opened 2022-10-21
feat: allow source tables other than `UnboundTable`

### LINKED ISSUE #393: Custom data source node doesn't work with selection
During testing of projection, we noticed that our custom data source node that doesn't work with the mapper counter logic here:

https://github.com/ibis-project/ibis-substrait/blob/main/ibis_substrait/compiler/translate.py#L743

This is because our internal data source only inherent from ibis `TableNode` but not `UnboundTable` (internally we have basically our own class for Data source node and we do not use ibis `UnboundTable` because at the time it seemed `UnboundTable` were used for different purpose).

This is from the discussion here:
https://github.com/ibis-project/ibis-substrait/blob/main/ibis_substrait/compiler/translate.py#L743

jcrist:
This change relaxes the check for source tables to allow any subclass of `PhysicalTable` that has a `.schema` attribute. While this isn't ideal in the long run, it's not terrible and is easy to do for now.

(Hopefully) fixes #393.

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -735,12 +735,16 @@ def selection(
         )
     ]
 
-    # TODO: there has to be a better way to get a list of unbound tables
+    # TODO: there has to be a better way to get a list of source tables
     # underlying an expression
-    unbound_tables = {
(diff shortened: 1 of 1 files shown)


icexelloss:
I tested internally and this works. If there is no objection I'd suggest we merge this patch and think more about long term solutions for this.

PR #403, opened 2022-10-29
chore: remove custom lockfile diffing and use our nixpkgs

cpcloud:
Cleans up lockfile checking and uses our pinned nixpkgs everywhere.


gforsyth (commented):
Thanks for this!

PR #411, opened 2022-11-02
feat: add support for ops.Log

### LINKED ISSUE #407: Add support for `Log`
Ibis-substrait currently supports `log10`, `log2`, and `ln`, but not `log`. AFAICT this is `logb` (https://github.com/substrait-io/substrait/blob/main/extensions/functions_logarithmic.yaml#L95) in substrait.

gforsyth:
if `base` isn't specified we default to the value of `e`

Resolves #407


cpcloud:
cc @icexelloss

Can you give this a try?

PR #413, opened 2022-11-02
feat: add support for manual ops.Clip

### LINKED ISSUE #409: Add support for `Clip`
`substrait` doesn't have a builtin method for this AFAICT, but the functionality may be composable from other builtin methods?

gforsyth:
This should be replaced if/when Substrait adds a `clip` scalar function definition

Resolves #409


cpcloud:
cc @icexelloss Clip is here, check it out

icexelloss:
Nice!

PR #415, opened 2022-11-03
feat: add more string mappings

richtia:
Adding more string mappings and fixing "Mean" which maps to avg in substrait.


cpcloud:
LGTM!

ibis_substrait/compiler/mapping.py
```diff
@@ -39,7 +39,7 @@
     "LPad": "lpad",
     "LStrip": "ltrim",
     "Max": "max",
-    "Mean": "mean",
+    "Mean": "avg",
```
cpcloud, line 42: @wjones127 Does Acero map this function correctly? I'm not sure whether it's called `mean` there.

wjones127: Yes. It's `avg` [in the yaml](https://github.com/substrait-io/substrait/blob/9d4c4d0dae179e7b1d91ee382f3e578af0015d08/extensions/functions_arithmetic.yaml#L830) and that is represented [in Acero here](https://github.com/apache/arrow/blob/cd61f81eafcb4d107e9b0b7548318513d4f4b3c3/cpp/src/arrow/engine/substrait/extension_set.cc#L915).

cpcloud: Great, then I'm merging this PR!

cpcloud (approved):

Issue #423, opened 2022-11-14
Emit Outputing mapping is incorrect for AsofJoin 


icexelloss:
Hi - I have found something wrong with Emit output mapping if I use asof join in my plan. The ibis-substrait what I used is in this branch: https://github.com/wjones127/ibis-substrait/tree/protobuf-3-19
Which includes a AsofJoin extension message:
https://github.com/wjones127/ibis-substrait/blob/protobuf-3-19/ibis_substrait/proto/extension_rels_pb2.py#L12

The code I have is internal, but it looks sth like this
```
t1 = ... #time, id, a
t2 = ... #time, id, b
joined = t1.asof_join(t2, on='time', by='id')
joined = joined[['time', 'id', 'a', 'b']]
```
When I execute the code with Acero, it crashes with a segment fault on this line
```
Thread 1 "/opt/ts/fuse/ar" received signal SIGSEGV, Segmentation fault.

arrow::engine::ProcessEmit<substrait::ProjectRel> (rel=..., no_emit_declr=..., schema=...) at /build/build/ext/public/apache/arrow/master/apache-arrow/cpp/src/arrow/engine/substrait/relation_internal.cc:78
```

After some debugging, it seems this is caused by the emit output mapping from the substrait plan, the emit mapping in my plan looks like this:

```
          emit {
            output_mapping: 6
            output_mapping: 7
            output_mapping: 8
            output_mapping: 9
          }
        }
```

After I hacked the ibis-substrait and forces it to produce the following emit mapping (based on information that @wjones127 told me), the query seems to work fine
```
          emit {
            output_mapping: 4
            output_mapping: 5
            output_mapping: 6
            output_mapping: 7
          }
        }
```

I am not sure how emit output mapping is supposed to work and where the problem is, please advice.

gforsyth:
Can you post the full substrait plan?

icexelloss:
```
 extension_uris {

  extension_uri_anchor: 1

}

relations {

  root {

    input {

      project {

        common {

          emit {

            output_mapping: 6

            output_mapping: 7

            output_mapping: 8

            output_mapping: 9

          }

        }

        input {

          extension_multi {

            inputs {

              read {

                base_schema {

                  names: "time"

                  names: "tid"

                  names: "return_raw_prev_1d_lag"

                  struct {

                    types {

                      timestamp_tz {

                        nullability: NULLABILITY_REQUIRED

                      }

                    }

                    types {

                      i32 {

[...]

PR #424, opened 2022-11-14
diff between protobuf-3-19 branch

wjones127:
* Rebuilt using protobuf 3.19
* Added asof join extension_rel protobuf. See `rebuild_protos.sh` for how that is included.
* Reverted fix to add enum options to function calls, so that we are compatible with anticipated changes in Substrait function calls.


rebuild_protos.sh
```diff
@@ -0,0 +1,37 @@
...
+# Clean compiled Python proto definitions
+rm -rf ./ibis_substrait/proto
+
+# Map the namespaces for the protos
+python proto_prefix.py "$proto_dir"/tmp substrait.ibis "$proto_dir"/substrait
+mv "$proto_dir"/substrait substrait.bak
+mv "$proto_dir"/tmp/substrait "$proto_dir"
+
+# Bring in the proto
+curl \
```
gforsyth, line 23: I think to upstream this we would want the extension_rels.proto to land in the upstream source tree so we aren't dependent on a particular fork to pull this down.

gforsyth: I'm also curious (not urgent) whether we can use the namespacing script to avoid the need for the `sed` commands below

ibis_substrait/compiler/translate.py
```diff
@@ -478,27 +481,12 @@ def value_op(
...
-
     # given the details of `op` -> function id
     return stalg.Expression(
         scalar_function=stalg.Expression.ScalarFunction(
             function_reference=compiler.function_id(expr),
             output_type=translate(expr.type()),
-            arguments=error_args
-            + [
+            arguments=[
                 stalg.FunctionArgument(value=translate(arg, compiler, **kwargs))
```
gforsyth, line 490: Happy to see these changes go -- once there's a release of the recently merged https://github.com/substrait-io/substrait/commit/bd29ea3b06391ae9018de851055db11075fd0758 we can upstream this change

ibis_substrait/compiler/translate.py
```diff
@@ -22,11 +22,14 @@
 import ibis.expr.schema as sch
 import ibis.expr.types as ir
 import toolz
+from google.protobuf.any_pb2 import Any as pb2_any
 from ibis import util
 from ibis.util import to_op_dag
 
+from ..proto import extension_rels_pb2 as extrel
```
gforsyth, line 29: This seems like a leftover since it's being redefined on line 32, that said, any proto thing that isn't namespaced into `substrait.ibis` can potentially collide with any C++ consumer because `protobuf`.  (It looks like this will be a non-issue, though)

ibis_substrait/compiler/translate.py
```diff
@@ -857,6 +845,74 @@ def join(
...
+
+    return stalg.Rel(
+        extension_multi=stalg.ExtensionMultiRel(
+            inputs=[
+                translate(op.left, compiler, **kwargs),
+                translate(op.right, compiler, **kwargs)
+            ],
+            detail=detail
+        )
+    )
```
gforsyth, line 880: This seems generally fine -- we'll have to think about how we want to handle translation rules that are linked to non-first-party extensions.

cpcloud: @icexelloss Can this live in your library that is using ibis instead of here? This particular implementation is highly specific to a particular choice of representation at the moment.

wjones127: This function was implemented by Bryce and I as a way to test whether the asof join extension relation could be used successfully. We don't have to include this.

rebuild_protos.sh
wjones127: As in, [this PR](https://github.com/substrait-io/substrait/pull/331)?

Or is there some other PR?

gforsyth: Ahh, I meant more that if it were living in `arrow` that we might pull it down from `apache/arrow`, but if it gets upstreamed into `substrait` that is even better.

ibis_substrait/compiler/translate.py
icexelloss: I don't have to include the translation rule but I think it's useful for reproducing issues. For example, issues like https://github.com/ibis-project/ibis-substrait/issues/423 can benefit from such rule so that we can at least run asof join with ibis and Acero in the open source

rebuild_protos.sh
wjones127: Got it. So perhaps long-term it will exist in Substrait, but that in the near term it might live as an extension in Arrow?

gforsyth: Right.  And in the near-term it may not make sense to have those proto definitions in ibis-substrait at all, they may belong in a standalone package.

gforsyth:
Hey @wjones127 -- I'm going to close this out for now -- the changes to error behavior are merged in and I think we've settled on `AsofJoin` living outside of this repository.

icexelloss:
Hello - following up here. Where should the `AsofJoin` substrait message live if not in ibis-substrait?

cpcloud:
It should be able to live anywhere really. Is there something preventing it from living within your own repo?

icexelloss:
My initial thought is that it will be better to live in ibis-substrait because
(1) one less repo
(2) makes it possible to integrate test asof join with ibis and Acero, both of which Voltron Data support and makes it easier to do integration test, benchmarks etc.

Why don't we want this in ibis-substrait?

cpcloud:
I think I am not particularly thrilled to have `AsOfJoin` in `ibis-substrait` because there is no `AsOfJoinRel` in the spec, making this a custom `ibis-substrait` extension.

I would be open to having it here if it existed upstream because it's a lot of work to maintain the custom protobufs (which also have to be maintained on the Acero side IIUC).

PR #448, opened 2022-12-20
feat: add initial support for ibis 4.x

gforsyth:
A lot of this is semi-gross hacks, but they should be relatively easy to unwind when we drop support for Ibis 3.x (whenever that happens).

Changes:

The biggest hurdle is that we no longer have the expr -> op -> expr ... pattern, it's just ops all the way down.

To maintain compatibility, any reference to something like `op.arg` is now written as `op.arg.op().to_expr()` -- this is gross, but functional. It's a very little bit of extra work, but it makes sure that we get the same object no matter the Ibis version.

Also, because now we're often passing around ops instead of exprs, or because the expr isn't strictly necessary for a bunch of dispatch rules, both `expr` and `compiler` are both keyword arguments now.

Besides that there are some minor shenigans to handle the nonexistence of `ops.CountStar` in ibis 3.x, and the renaming / rearrangement of the graph utilities.

All in all, there are only a few explicit 4.x vs 3.x branches.

Also, need to add ibis 4.x to CI and then make sure we test both versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,9 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
+        ibis-version:
+          - "3.2"
+          - "4"
 
     steps:
       - uses: actions/checkout@v3
@@ -102,7 +105,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - run: pip3 install poetry
-      - run: poetry install
+      - run: poetry install --sync
+      - run: poetry run pip install ibis-framework==${{ matrix.ibis-version }}
       - run: poetry run pytest
   dry-run-release:
     runs-on: ubuntu-latest
--- a/shell.nix
+++ b/shell.nix
@@ -11,7 +11,6 @@ let
     git
     jq
     niv
-    nix-linter
     nixpkgs-fmt
     prettierTOML
     protobuf3_20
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -102,7 +105,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - run: pip3 install poetry
-      - run: poetry install
+      - run: poetry install --sync
+      - run: poetry run pip install ibis-framework==${{ matrix.ibis-version }}
```
gforsyth, line 109: I'm not proud of this but it works.

shell.nix
```diff
@@ -11,7 +11,6 @@ let
     git
     jq
     niv
-    nix-linter
```
gforsyth, line 14: Temporary (and perhaps unnecessary) but I was getting

```
error: Package ‘nix-linter-0.2.0.4’ in /nix/store/hi1s6s43602ypv0vqi8gy36g01gw6l6s-nixpkgs-src/pkgs/development/tools/analysis/nix-linter/default.nix:23 is marked as broken, refusing to evaluate.
```

cpcloud: This is fine, we'll move to `statix` + `deadnix`.

cpcloud (approved):

cpcloud:
LGTM, merge when you're ready. This will cut a sweet new feature release!

gforsyth:
@cpcloud -- are there any changes required to the CI required status checks required?

cpcloud:
I'm disabling the "branch must be up-to-date" requirement.

PR #465, opened 2023-01-13
refactor: clean up output mapping in ops.Selection dispatch

gforsyth:
This code was a bit spaghetti and hard to follow, especially with the extra branching for multiple versions of Ibis.  Cleaning it up a little bit to make it easier to follow.

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -775,42 +775,7 @@ def selection(
             )
         )
 
-    # TODO: there has to be a better way to get a list of source tables
-    # underlying an expression
-    # TODO: settle on a better source table definition than "PhysicalTable with
-    # a schema"
-    if IBIS_4:
-        # projection / emit
-        selections = [
-            col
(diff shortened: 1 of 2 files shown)


ibis_substrait/compiler/translate.py
```diff
@@ -860,6 +828,48 @@ def selection(
...
+
+
+def _find_parent_tables(op: ops.Selection) -> set[ir.Table]:
+    # TODO: settle on a better source table definition than "PhysicalTable with
+    # a schema"
+    if IBIS_4:
+        source_tables = {
+            t
+            for t in toposort(op).keys()
+            if isinstance(t, ops.PhysicalTable) and hasattr(t, "schema")
```
westonpace, line 861: What's a `PhysicalTable`?  I thought that meant "table on disk" / "source node" but I suspect I'm wrong?

gforsyth: It's weirdly named -- it's the parent class of `UnboundTable`, `NamedTable`, `DatabaseTable` and others.  

Here it effectively fills in for "anything that probably has a `schema`" and then we double-check that.

PR #466, opened 2023-01-13
fix(emit): check all rel types for existing output mapping

### LINKED ISSUE #423: Emit Outputing mapping is incorrect for AsofJoin

gforsyth:
Fixes #423 

@icexelloss can you try this out on your asof_join extension?

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -777,13 +777,32 @@ def selection(
 
     if selections := _get_selections(op):
 
-        if relation.project.common.ListFields():
-            # if there is already an `emit` in RelCommon then we're stacking
-            # projections and we need to update the output_mapping to refer to
-            # the number of fields present in the most recent emit
-            mapping_counter = itertools.count(
-                len(relation.project.common.emit.output_mapping)
+        rels = [
+            (attr, getattr(relation, attr))
+            for attr in (
+                "aggregate",
+                "cross",
+                "extension_leaf",
+                "extension_multi",
+                "extension_single",
+                "fetch",
+                "filter",
+                "join",
+                "project",
+                "read",
+                "set",
+                "sort",
             )
+        ]
+        for relname, rel in rels:
+            if relname == "aggregate" and rel.measures:
+                mapping_counter = itertools.count(
+                    len(rel.measures) + len(rel.groupings)
+                )
+                break
+            elif output_mapping := rel.common.emit.output_mapping:
+                mapping_counter = itertools.count(len(output_mapping))
+                break
         else:
             source_tables = _find_parent_tables(op)
             mapping_counter = itertools.count(sum(len(t.schema) for t in source_tables))
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -33,11 +33,15 @@ IBIS_4 = False
 try:
     from ibis.util import to_op_dag
 
-    # Bit of a hack -- there is no ops.CountStar in Ibis 3.x but to register it
-    # for 4.x below, it can't be undefined here.
+    # There is no ops.CountStar in Ibis 3.x but to register it for 4.x below, it
+    # can't be undefined here.
     # We remap it to ops.Count just to avoid an attribute error, it will never
     # be used to route in Ibis 3.x because it doesn't exist.
     ops.CountStar = ops.Count
+    # ops.ValueOp renamed to ops.Value in Ibix 3.2
+    # We manually add ops.Value here for Ibis 3.0 compatibility
+    if hasattr(ops, "ValueOp"):
+        ops.Value = ops.ValueOp
 except ImportError:
     import warnings
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,21 +147,15 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
 
-      - name: install nix
-        uses: cachix/install-nix-action@v18
+      - uses: actions/setup-python@v4
         with:
-          nix_path: nixpkgs=./nix
-          extra_nix_config: |
-            access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
+          python-version: "3.x"
 
-      - name: setup cachix
-        uses: cachix/cachix-action@v12
-        with:
-          name: ibis-substrait
-          extraPullNames: nix-community,poetry2nix
+      - name: install poetry
+        run: pip3 install poetry
 
       - name: verify lockfile
-        run: nix run 'nixpkgs#poetry' -- lock --check
+        run: poetry lock --check
   release:
     runs-on: ubuntu-latest
     if: ${{ github.event_name != 'pull_request' }}
--- a/ibis_substrait/tests/compiler/test_compiler.py
+++ b/ibis_substrait/tests/compiler/test_compiler.py
@@ -251,6 +251,36 @@ def test_emit_nested_projection_output_mapping(compiler):
     assert result.project.common.emit.output_mapping == [3, 4]
 
 
+def test_aggregate_project_output_mapping(compiler):
+    t = ibis.table(
+        [
+            ("a", "int64"),
+            ("b", "int64"),
+            ("c", "int64"),
+            ("d", "int64"),
+            ("e", "int64"),
+        ],
+        name="t",
+    )
+
+    expr = t.group_by(["a", "b"]).aggregate([t.c.sum().name("sum")]).select("b", "sum")
+
+    result = translate(expr, compiler)
+    assert result.project.common.emit.output_mapping == [3, 4]
+
+    # Select 3 of 5 columns in base table to make sure output mapping reflects
+    # a count from the outputs of the aggregation ('a', 'b', and 'sum')
+    expr = (
+        t.select("a", "b", "c")
+        .group_by(["a", "b"])
+        .aggregate([t.c.sum().name("sum")])
+        .select("b", "sum")
+    )
+
+    result = translate(expr, compiler)
+    assert result.project.common.emit.output_mapping == [3, 4]
+
+
 def test_ibis_schema_to_substrait_schema():
     input = ibis.schema(
         [


ibis_substrait/compiler/translate.py
```diff
@@ -777,13 +777,27 @@ def selection(
...
+                "extension_leaf",
+                "extension_multi",
+                "extension_single",
+                "fetch",
+                "filter",
+                "join",
+                "project",
+                "read",
+                "set",
+                "sort",
```
gforsyth, line 794: I would be more clever about this, but there's already enough "cleverness" in `protobuf` that I opted for simple.

westonpace:
Hmm...I'm still not sure this satisfies the `read -> aggregation -> project` case.  I would assume that `aggregation` wouldn't bother specifying an output mapping since the default is probably just fine.  But the default is not "propagate output mapping from the input"

gforsyth:
> Hmm...I'm still not sure this satisfies the `read -> aggregation -> project` case. I would assume that `aggregation` wouldn't bother specifying an output mapping since the default is probably just fine. But the default is not "propagate output mapping from the input"

Yes, you're right.  

I guess we can either explicitly `emit` the aggregate outputs, or count the `selection`s in `groupings` and `measures` (only if we end up doing a `project` on top of an `aggregation`

gforsyth:
Ok, I pushed a fix and added a test for the `read -> aggregate -> project` case.

There may be other relations that require a bit of special handling, but hopefully this moves the ball forward.

icexelloss:
@gforsyth Check in here - this is done and you want me to test it with my acero integration test right?

ibis_substrait/compiler/translate.py
```diff
@@ -777,13 +777,32 @@ def selection(
...
             )
+        ]
+        for relname, rel in rels:
+            if relname == "aggregate" and rel.measures:
```
icexelloss, line 798: Can you comment on why do we handle aggregate differently?

gforsyth: see above in https://github.com/ibis-project/ibis-substrait/pull/466#issuecomment-1382311006

ibis_substrait/compiler/translate.py
```diff
@@ -777,13 +777,32 @@ def selection(
 
     if selections := _get_selections(op):
 
-        if relation.project.common.ListFields():
-            # if there is already an `emit` in RelCommon then we're stacking
-            # projections and we need to update the output_mapping to refer to
-            # the number of fields present in the most recent emit
-            mapping_counter = itertools.count(
-                len(relation.project.common.emit.output_mapping)
+        rels = [
```
icexelloss, line 780: Why do need to have a list of predefined node types instead of doing it for all the node types? (I thought emit output_mapping is a thing that should work for all node types?)

icexelloss:
@gforsyth Can you remind me how to create source tarball for ibis-substrait?

ibis_substrait/compiler/translate.py
gforsyth: it was just simpler than querying all of the attributes of the object (also because google does weird things in protobuf and creates attributes that don't exist)

gforsyth:
> @gforsyth Can you remind me how to create source tarball for ibis-substrait?

zip file ok? https://github.com/gforsyth/ibis-substrait/archive/refs/heads/gil/output_mapping_extension.zip

icexelloss:
Nvm I found my notes about this - "Build the source tarball with poetry build"

icexelloss:
I think the new output mapping looks correct now - I still need to test it with changes on the Acero side (hopefully later Today).

Will update here

snadukudy:
> I think the new output mapping looks correct now - I still need to test it with changes on the Acero side (hopefully later Today).
> 
> Will update here

great news @icexelloss, please keep us posted on your testing. thank you, Sri

icexelloss:
Unfortunately, my results shows it's still not quite right. Here is how the plan looks like:

```
extension_uris {
  extension_uri_anchor: 1
}
relations {
  root {
    input {
      project {
        common {
          emit {
            output_mapping: 4
            output_mapping: 5
            output_mapping: 6
            output_mapping: 7
          }
        }
        input {
          extension_multi {
            common {
              emit {
                output_mapping: 0
                output_mapping: 1
                output_mapping: 2
                output_mapping: 5
              }
            }
            inputs {
              project {
                common {
                  emit {
                    output_mapping: 3
                    output_mapping: 4
                    output_mapping: 5
                  }
                }
                input {
                  project {
                    common {
                      emit {
                        output_mapping: 3
                        output_mapping: 4
                        output_mapping: 5
                      }
                    }
                    input {
                      read {
                        base_schema {
                          names: "time"
                          names: "tid"
                          names: "price"
                          struct {
                            types {
                              timestamp_tz {
                                nullability: NULLABILITY_REQUIRED
                              }
                            }
                            types {
                              i32 {
                                nullability: NULLABILITY_NULLABLE
                              }
                            }
                            types {
                              fp64 {
                                nullability: NULLABILITY_NULLABLE
                              }
                            }
                            nullability: NULLABILITY_REQUIRED
                          }
                        }
                        named_table {
                          names: "smooth:/research/group/tsmoto/bamboo/ci/test1?begin=1546300800000000000&end=1547078400000000000&columns=tid,price"
                        }
                      }
                    }
                    expressions {
                      selection {
                        direct_reference {
                          struct_field {
                          }
                        }
                        root_reference {
                        }
                      }
                    }
                    expressions {
                      selection {
                        direct_reference {
                          struct_field {
                            field: 1
                          }
                        }
                        root_reference {
                        }
                      }
                    }
                    expressions {
                      selection {
                        direct_reference {
                          struct_field {
                            field: 2
                          }
                        }
                        root_reference {
                        }
                      }
                    }
                  }
                }
                expressions {
                  selection {
                    direct_reference {
                      struct_field {
                      }
                    }
                    root_reference {
                    }
                  }
                }
                expressions {
                  selection {
                    direct_reference {
                      struct_field {
                        field: 1
                      }
                    }
                    root_reference {
                    }
                  }
                }
                expressions {
                  selection {
                    direct_reference {
                      struct_field {
                        field: 2
                      }
                    }
                    root_reference {
                    }
                  }
                }
              }
            }
            inputs {
              project {
                common {
                  emit {
                    output_mapping: 3
                    output_mapping: 4
                    output_mapping: 5
                  }
                }
                input {
                  project {
                    common {
                      emit {
                        output_mapping: 3
                        output_mapping: 4
                        output_mapping: 5
                      }
                    }
                    input {
                      read {
                        base_schema {
                          names: "time"
                          names: "tid"
                          names: "price"
                          struct {
                            types {
                              timestamp_tz {
                                nullability: NULLABILITY_REQUIRED
                              }
                            }
                            types {
                              i32 {
                                nullability: NULLABILITY_NULLABLE
                              }
                            }
                            types {
                              fp64 {
                                nullability: NULLABILITY_NULLABLE
                              }
                            }
                            nullability: NULLABILITY_REQUIRED
                          }
                        }
                        named_table {
                          names: "smooth:/research/group/tsmoto/bamboo/ci/test1_daily?begin=1546300800000000000&end=1547078400000000000&columns=tid,price"
                        }
                      }
                    }
                    expressions {
                      selection {
                        direct_reference {
                          struct_field {
                          }
                        }
                        root_reference {
                        }
                      }
                    }
                    expressions {
                      selection {
                        direct_reference {
                          struct_field {
                            field: 1
                          }
                        }
                        root_reference {
                        }
                      }
                    }
                    expressions {
                      selection {
                        direct_reference {
                          struct_field {
                            field: 2
                          }
                        }
                        root_reference {
                        }
                      }
                    }
                  }
                }
                expressions {
                  selection {
                    direct_reference {
                      struct_field {
                      }
                    }
                    root_reference {
                    }
                  }
                }
                expressions {
                  selection {
                    direct_reference {
                      struct_field {
                        field: 1
                      }
                    }
                    root_reference {
                    }
                  }
                }
                expressions {
                  selection {
                    direct_reference {
                      struct_field {
                        field: 2
                      }
                    }
                    root_reference {
                    }
                  }
                }
              }
            }
            detail {
              type_url: "type.googleapis.com/arrow.substrait_ext.AsOfJoinRel"
              value: "\n\022\n\006\022\004\n\002\022\000\022\010\022\006\n\004\022\002\010\001\n\022\n\006\022\004\n\002\022\000\022\010\022\006\n\004\022\002\010\001\020\200\240\361\302\2614"
            }
          }
        }
        expressions {
          selection {
            direct_reference {
              struct_field {
              }
            }
            root_reference {
            }
          }
        }
        expressions {
          selection {
            direct_reference {
              struct_field {
                field: 1
              }
            }
            root_reference {
            }
          }
        }
        expressions {
          selection {
            direct_reference {
              struct_field {
                field: 2
              }
            }
            root_reference {
            }
          }
        }
        expressions {
          selection {
            direct_reference {
              struct_field {
                field: 3
              }
            }
            root_reference {
            }
          }
        }
      }
    }
    names: "time"
    names: "tid"
    names: "price"
    names: "price_right"
  }
}
```

My code looks like:
```
       dt = self.basic_smooth_table[['tid', 'price']]
       dt1 = self.basic_daily_smooth_table[['tid', 'price']]
       join = dt.asof_join(dt1, by=["tid"], tolerance=-30 * bb.time.MINUTES)
       join = join[['tid', 'price', 'price_right']]
```

Result:
```
                                tid               price  price_right
time                                                               
2019-01-02 05:00:00+00:00       463 2019-01-02 05:00:00          463
2019-01-02 05:00:00+00:00      1829 2019-01-02 05:00:00         1829
2019-01-02 05:00:00+00:00      5803 2019-01-02 05:00:00         5803
2019-01-02 05:00:00+00:00      6466 2019-01-02 05:00:00         6466
2019-01-02 05:00:00+00:00      6658 2019-01-02 05:00:00         6658
...                             ...                 ...          ...
2019-01-09 21:00:00+00:00  36763521 2019-01-09 21:00:00     36763521
2019-01-09 21:00:00+00:00  39701460 2019-01-09 21:00:00     39701460
2019-01-09 21:00:00+00:00  57250345 2019-01-09 21:00:00     57250345
2019-01-09 21:00:00+00:00  58524014 2019-01-09 21:00:00     58524014
2019-01-09 21:00:00+00:00  58665129 2019-01-09 21:00:00     58665129
```

Acero version: https://github.com/westonpace/arrow/commits/feature/bamboo-demo
Ibis-substrait version: https://github.com/gforsyth/ibis-substrait/commits/gil/output_mapping_extension

gforsyth:
@icexelloss -- can you clarify what isn't right about it?

Are you getting columns you don't expect (or is it missing expected outputs?) or are the computed results incorrect?

icexelloss:
Yes - the result table column index is wrong. I think the “price” is the
“time” column and the “price_right” is the “tux” column

icexelloss:
Typo: I meant the “price_right” is the “tid” column

gforsyth:
@icexelloss -- can you re-run but leave off the final selection?  I'd like to see some sample results and the substrait plan for "just" the `asof_join`, like the result and plan for

```python
       dt = self.basic_smooth_table[['tid', 'price']]
       dt1 = self.basic_daily_smooth_table[['tid', 'price']]
       join = dt.asof_join(dt1, by=["tid"], tolerance=-30 * bb.time.MINUTES)
```

and can you post the Ibis expression repr for those `dt` objects?

icexelloss:
Without the final project the substrait plan looks exactly like the one with the projection - except for that it doesn't have the projection node after the asof join of course. It also gives the correct result. 

I am not sure the ibis expr repr is too interesting - what are you trying to see from it?

icexelloss:
Anyway here is the ibis tree
```
(Pdb) execution_plan

r0 := IbisUnboundSmoothURITableNode

    name:

      smooth:/research/group/tsmoto/bamboo/ci/test1_daily

    region:

      None

    snapshot:

     None

    _schema:

      time  !timestamp('UTC')

      tid   int32

      price float64

   

 

r1 := IbisUnboundSmoothURITableNode

    name:

      smooth:/research/group/tsmoto/bamboo/ci/test1

    region:

      None

    snapshot:

      None

    _schema:

      time  !timestamp('UTC')

      tid   int32

      price float64

   

 

r2 := Selection[r0]

  selections:

    time:  r0.time

    tid:   r0.tid

    price: r0.price

 

r3 := Selection[r1]

  selections:

    time:  r1.time

    tid:   r1.tid

    price: r1.price

 

r4 := IbisAsOfJoinNode:

  time        !timestamp('UTC')

  tid         int32

  price       float64

  price_right float64

 

Selection[r4]

  selections:

    time:        r4.time

    tid:         r4.tid

    price:       r4.price

    price_right: r4.price_right
```

gforsyth:
Thanks @icexelloss -- I'm just trying to get a handle on how many columns the operation might return before selection / projection.

I'd like to try one more thing.  Can you mutate the `output_mapping` of your `extension_multi` node so that it reads `[6, 7, 8, 11]` instead of `[0, 1, 2, 5]`?

If the variable containing the proto bytes was called `plan`, this would look something like

```python
plan.relations[0].root.input.project.input.extension_multi.common.emit.output_mapping = [6, 7, 8, 11]
```

icexelloss:
Sure I can try - why do you want to try this and which one is correct by substrait definition?

icexelloss:
After I changed to `[6, 7, 8, 11]` I got "pyarrow.lib.ArrowInvalid: Out of bounds emit index 6"

gforsyth:
> Sure I can try - but why do you want to try this and which one is correct by substrait definition/spec?

I thought that what I was suggesting was correct, but if it's giving an out-of-bounds I'm probably wrong

icexelloss:
> > Sure I can try - but why do you want to try this and which one is correct by substrait definition/spec?
> 
> I thought that what I was suggesting was correct, but if it's giving an out-of-bounds I'm probably wrong

Not necessarily that you are wrong if it gives out-of-bounds error - it could also be that Acero is not handling this correctly so I am trying to understand the correct "plan" by substrait spec

gforsyth:
My understanding of how `emit` works is that the count for the `output_mapping` should also include the field references from the incoming stream / data / columns.

For an inner join (which is obviously different, but for an example), input columns

time
tid_L
price_L
time
tid_R
price_R

get stacked together and any _outputs_ of the `inner_join` start at index `6`, because the inputs are currently occupying spots [0-5].

so my thinking was that the outputs of the `asof_join` would also start at `6` -- although thinking about it again, I would expect the `output_mapping` to be `[6, 7, 8, 9]`

westonpace:
An inner join has no `outputs`.  It simply concatenates two inputs.

icexelloss:
I see. Well I don't understand emit well enough yet so let's ping @westonpace to decide on what the right emit output should be here. 

I thought we are out of the weeds of what the right "plan" should look like here but I guess we are circling back yet again.

westonpace:
`read`, `project` and `aggregate` are the only nodes today (off the top of my head) that create new columns.

icexelloss:
For @westonpace:

If (still big if) the substrait plan looks like, then my current theory is that this patch https://github.com/westonpace/arrow/commit/c5ac1260243bc2d8522d892d4a1afbd3e9524caa doesn't handle the case when the asof join has a emit output. I spent quite a bit time debugging the code but it is a bit confusing to me (I am not sure if we still need the field_output_indices if we have the emit output - seems to be they are the same thing). I am probably done for the day, but just want to share my findings and thoughts.

westonpace:
Assuming an input of:

```
time_L
tid_L
price_L
time_R
tid_R
price_R
```

The output of both join and asof-join (in Substrait) will be the same:

```
0: time_L
1: tid_L
2: price_L
3: time_R
4: tid_R
5: price_R
```

I agree with Acero (naturally :) that 6 is out of bounds.

gforsyth:
Ok, @westonpace's comment has clarified things for me -- in my example above, the output_mapping starts at 6 because we're going (in my head) a projection after the inner join.

In the case of `asof_join`, we're doing a projection after the join, the join has 4 outputs, according to the `emit`, and then we project on top of that and should get the 4 new columns created, with indices [4, 5, 6, 7]

icexelloss:
@westonpace @gforsyth have we agreed on what the emit output should look like? 

> and then we project on top of that and should get the 4 new columns created, with indices [4, 5, 6, 7]
This means the emit output is correct in the plan's I provided or no?

icexelloss:
@westonpace I have one more question about the plan that I generated - 

For the projection node, it seems that emit output starting index is "number of input columns", i.e.,
if I have 3 inputs and I select all of them, the emit output is `[3, 4, 5]`

For the asof join node, it seems that emit output starting index is "0", i.e.,
if I have 3 inputs from left, 3 inputs from right, and output 4 columns (time, tid is from right is de duplicated), the emit output if `[0, 1, 2, 5]`

What is the rule what decides "what is a starting emit index" for a node?

westonpace:
First, we should probably note that all of these project nodes are pointless, though harmless.  In Substrait, a project node's only purpose is to add columns.  It cannot reorder or reduce the number of columns.  So the default output of a project node is `[all input columns, one column per expression]`.

So the Substrait relation `project(field(0), field(1), field(2), field(3))` has an output schema (assuming there are 4 input fields) of `f0, f1, f2, f3, f0, f1, f2, f3` and the emit `[4, 5, 6, 7]` and the emit `[0, 1, 2, 3]` would give the exact same result (as would removing the project relation entirely).

If this all seems excessive or silly it's because it is.  I imagine what is going on here is that Ibis is used to thinking of proejctions as "selections" where there is one output column per expression.  In this case, since the output of the project node is `[all input columns, one column per expression]` you will always get a "selection" if you just take the second half of the output `[one column per expression]`.  This is why a "selection" is represented as a "project" with an emit starting at the # of input fields.

icexelloss:
I think I have good news and bad news.

Good news is that I think I managed to get the correct result with this patch with the Acero code from https://github.com/westonpace/arrow/commits/feature/bamboo-demo

Bad news is that some of out existing test broke because I think ibis-substrait no longer works with ibis 3? (I am seeing:
```
Error in translate.py:519

AttributeError: module 'ibis.expr.operations' has no attribute 'Value'
```
@gforsyth Did we drop ibis 3 support?

icexelloss:
```
So the Substrait relation project(field(0), field(1), field(2), field(3)) has an output schema (assuming there are 4 input fields) of f0, f1, f2, f3, f0, f1, f2, f3 and the emit [4, 5, 6, 7] and the emit [0, 1, 2, 3] would give the exact same result (as would removing the project relation entirely).
```
Thanks @westonpace. Why would projection duplicate the 4 inputs to `f0, f1, f2, f3, f0, f1, f2, f3`? Is that a substrait thing for just projection? Or does other node follow this rule? (prepend inputs to output schema?)

gforsyth:
> @gforsyth Did we drop ibis 3 support?

Not intentionally -- if you can put up a PR with the failing test I can fix whatever we missed.

gforsyth:
> Not intentionally -- if you can put up a PR with the failing test I can fix whatever we missed.

nevermind, I think I know what's wrong.  I'll push a patch up to this branch

icexelloss:
Thanks - let me try this now.

Is https://github.com/ibis-project/ibis-substrait/pull/466/commits/5892a7baba7a07fd86043cf66691e8c383266b18 sth we can push to the default/master along with this patch?

gforsyth:
yeah, I'm happy to merge this -- just adding one more small fix for ibis 3.x

icexelloss:
FWIW, I tested my use with ibis 3.2 and it seems OK. The issue appears to be that ops.ValueOp is renamed to ops.Value between 3.0 and 3.2 (IMHO that should probably be a 4.x change since removing ops.ValueOp is quite a breaking change)

icexelloss:
> yeah, I'm happy to merge this -- just adding one more small fix for ibis 3.x

Great! Let me test this now

icexelloss:
@gforsyth tests on my side passed! I think we can go ahead to merge this to default

cpcloud (approved):
SHIP IT 🚢 🚢

snadukudy:
Awesome! thank you team :)

westonpace:
> Thanks @westonpace. Why would projection duplicate the 4 inputs to f0, f1, f2, f3, f0, f1, f2, f3? Is that a substrait thing for just projection? Or does other node follow this rule? (prepend inputs to output schema?)

The only other nodes that create new columns are read (which has no inputs) and the aggregate node.  It is impossible to include the input columns in the aggregate case (since they likely have different length) and so the output for the aggregate node is just the new columns.

All other nodes pass their inputs through (e.g. filter, union) or combine their inputs (join, asof join).

icexelloss:
So to summarize the output schema
Projection: input + output
Read: output
Aggregate: output
Filter, union (pass through): input
Join, asof join (combine): input

Is that the correct understanding?

westonpace:
Yes.  You can add `sort`, `fetch`, and `set`, and `write` to `(pass through)` and you can add `Cross Product` to `(combine)`.  Then I think you've covered the full set of logical operators that exist today.

icexelloss:
This makes sense. Project is the one that kind of stand out and confused me.

PR #477, opened 2023-01-18
build: add build.py script to generate protobuf code at install/build time

cpcloud:
This PR should support custom compilation of protobuf sources by way of the `PROTO_DIR` environment variable


gforsyth:
We can probably automate _some_ of this (not that I'm saying we _should_) but we need to at least document that if using `PROTO_DIR` then the following need to already be installed:
`mypy_protobuf`, `protoletariat`, `protobuf`, `buf` (not available via common python channels)

we _can_ add those (except `buf`) to `build-system-requires` but that, I think, will add things that are unnecessary for most users

cpcloud:
In theory we can use something other than `buf` here to avoid that being a necessary dependency.

gforsyth (approved):
I've tested this locally and while we need to add docs, that doesn't need to hold this up.

PR #510, opened 2023-02-15
feat(extensions): add extension URIs to plans

gforsyth:
Ok, this is take 2 on adding in proper extension support.

Substrait wants scalar function calls to list a URI that points to a YAML file that defines the name of the scalar function, as well as any options it can take, and also which types are valid inputs.

This turns out to be a bit of a pain to handle automatically -- the function `add`, for instance, is defined in two separate YAML files, one that specifies behavior for ints and floats, the other for decimals.

So here, we add in a `register` function that parses a substrait `YAML` extension file and builds up a mapping of the form:

`map[function name][('tuple', 'of', 'input', 'types')]`

and the value associated with that entry is an object that has the options, URI, etc, associated with that particular function call.

To add to the fun, we can't just pass in the `input` types as we get them, because there are loads of edge cases, so I've put together the following (horrible) process:

Check if function takes arguments of `any` type.  Some functions accept any input, so we need to see if that's an option -- we still check here that the number of arguments we're trying to pass in is correct.

If the function doesn't take `any` type, then we attempt to normalize (bring to parity?) the input datatypes.  If we're trying to add two columns and one is `int32` and the other is `int64`, we insert a `cast` on the `int32` before passing it into the scalar function dispatch.
Substrait only defines these scalar functions for equal types, it doesn't allow mixing types as inputs.  Some consumers might automagically upcast for us, but we try to be a good citizen. 

There's a fun sub-edge case here for things like `substring`, which expects something like `('varchar', 'i32', 'i32')` -- there are a limited number of string ops that expect integer arguments, so those are dispatched on explicitly to cast those numeric argument(s) to `i32`.

Now when we look up the extension signature, in the case that our inputs exactly match what the function takes, we're done, e.g. `map['add'][('i64', 'i64')]`
 
If we don't have a match on signature, we check if the extension is variadic.  If it's defined as `variadic`, then it will only have a single value-per-type listed, so we might be trying to look up `('boolean', 'boolean')` but the function will be parsed as `('boolean',)`.
So if we've tried everything else, we trim the signature and try that -- then check to make sure that the `variadic` flag is `True` on the returned object if there's a match.

If none of this works, we error out.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,4 @@ flake.lock linguist-generated=true
 poetry.lock linguist-generated=true
 ibis_substrait/proto/** linguist-generated=true
 proto/** linguist-generated=true
+ibis_substrait/extensions/** linguist-generated=true
--- a/gen-protos.sh
+++ b/gen-protos.sh
@@ -7,15 +7,22 @@ substrait_protos="${1:-${PROTO_DIR}}"
 set -u
 
 proto_dir=./proto
+extension_dir=./ibis_substrait/extensions
 
 mkdir -p "$proto_dir"
+mkdir -p "$extension_dir"
 chmod u+rwx "$proto_dir"
+chmod u+rwx "$extension_dir"
 rm -r "$proto_dir"
+rm -r "$extension_dir"
 
 cp -fr "$substrait_protos" "$proto_dir"
+cp -fr "$substrait_protos/../extensions" "$extension_dir"
 
 find "$proto_dir" -type d -exec chmod u+rwx {} +
 find "$proto_dir" -type f -exec chmod u+rw {} +
+find "$extension_dir" -type d -exec chmod u+rwx {} +
+find "$extension_dir" -type f -exec chmod u+rw {} +
 
 rm -rf ./ibis_substrait/proto
 
--- a/ibis_substrait/compiler/core.py
+++ b/ibis_substrait/compiler/core.py
@@ -7,13 +7,17 @@ from typing import Any, Hashable, Iterator
 
 import google.protobuf.message as msg
 import ibis.expr.datatypes as dt
+import ibis.expr.operations as ops
 import ibis.expr.types as ir
 
+from ibis_substrait.compiler.mapping import (
+    IBIS_SUBSTRAIT_OP_MAPPING,
+    IBIS_SUBSTRAIT_TYPE_MAPPING,
+    _extension_mapping,
+)
 from ibis_substrait.proto.substrait.ibis import algebra_pb2 as stalg
 from ibis_substrait.proto.substrait.ibis import plan_pb2 as stp
 from ibis_substrait.proto.substrait.ibis.extensions import extensions_pb2 as ste
-
-from .mapping import IBIS_SUBSTRAIT_OP_MAPPING
 
 
 def which_one_of(message: msg.Message, oneof_name: str) -> tuple[str, Any]:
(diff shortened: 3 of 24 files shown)


westonpace:
> However, Substrait doesn't allow mixing types as inputs, it expects you to announce what the type should be and rely on the consumer to upcast as necessary.
So, if we still haven't found a valid signature, we check the precedence of the arguments we have, cast up as possible, then try again.

This is unclear to me.  The first sentence is slightly off:

> rely on the consumer to upcast as necessary.

No, it expects for the cast to be explicit.  In other words, `add("my_i32_field", literal(5, int8()))` is not a valid plan but `add("my_i32_field", cast(literal(5, int8())))` is (although, in that case, you might as well do `literal(5, int32())`).

However, your second sentence is:

> we check the precedence of the arguments we have, cast up as possible, then try again.

which seems to be the correct thing to do but doesn't match the first sentence.

westonpace:
By the way, if it's any consolation, the cast behavior you are describing sounds exactly the implicit casting algorithm we have in Acero so this is probably always happening _somewhere_ it just isn't your responsibility in the other backends (happy to pass the buck :laughing:)

westonpace:
As I understand it this generates something like:

```
{
      "extensionUriAnchor": 1,
      "uri": "./functions_boolean.yaml"
}
```

That doesn't seem to be a URI.  However, this is pretty close to the way that Isthmus handles URIs as well (and I'm pretty sure DuckDb just leaves it blank).  So now I will bore you with my little rant.

One could argue that the "canonical" URI is `https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml` (stable URI, unstable content).  One could also argue that it is `https://github.com/substrait-io/substrait/blob/v0.25.0/extensions/functions_boolean.yaml` (unstable URI, stable content).  Or you could go for `https://raw.githubusercontent.com` instead of `https://github.com` (less friendly looking URI but yields an actual file).  Or you could argue that the URI should be `https://substrait.io/extensions/functions_boolean.yaml` (doesn't actually resolve but it could be made to do so and you would remove a dependency on github hosting).

At the moment, no one has actually cared enough to solve this problem.  Assuming you also do not, then I would request you use either `/functions_boolean.yaml` or `functions_boolean.yaml` as the URI (and NOT `./functions_boolean.yaml`) which will allow you to be more consistent with the other producers (if everyone does this then maybe we just someday erase the fact that this was ever intended to be a URI).

westonpace:
Acero should recognize `https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml`, `/functions_boolean.yaml` and `functions_boolean.yaml`.  It will even tolerate the empty string (in that case we fallback to name-only matching).

No other consumer that I am aware of has properly implemented function mapping.

gforsyth:
Thanks @westonpace !

I'll work on clearing up the language above, since I imagine I'll need to refer back to this PR at some point.

In re: casting -- we're currently only casting as a check that we can use the function that we're trying to (so that we'll fail out if we try to add a `varchar` and an `int8`), but I think that I should update that to also add the actual `cast` to the plan.

> At the moment, no one has actually cared enough to solve this problem. Assuming you also do not, then I would request you use either /functions_boolean.yaml or functions_boolean.yaml as the URI (and NOT ./functions_boolean.yaml) which will allow you to be more consistent with the other producers (if everyone does this then maybe we just someday erase the fact that this was ever intended to be a URI).

I'm trying really hard to not get nerdsniped by this.
I'm going to work on casting a bit first, then I'll circle back.

I like the idea of an actual URI pointing to either github or subtrait.io, especially if we have users who might want to have custom extension files.

Issue #531, opened 2023-03-10
Add PyArrow <-> ibis-substrait integration tests


amoeba:
It would be beneficial to this project if the type of testing we have done manually in PRs like https://github.com/ibis-project/ibis-substrait/pull/414 could be automated. I think everything that would be needed to do this exists now.

As an example of what these kinds of tests could look like, I've been using [this Python script](https://gist.github.com/amoeba/a5a62a75301eb42b10ef21d70e7086b5) to run Ibis expressions against Arrow Tables via Substrait.

For bringing in a recent PyArrow build with Substrait support enabled, CI could install pyarrow with `pip install --extra-index-url https://pypi.fury.io/arrow-nightlies --pre pyarrow`. The conda `arrow-nightlies` channel also has fairly recent (but not as recent as gemfury) builds.

PR #532, opened 2023-03-10
feat: Add WIP basic pyarrow integration test

amoeba:
Ref #531

Hey @gforsyth, this is a bit WIP but `pytest` runs these if `pyarrow` and `pyarrow.substrait` are installed (like via `pip install --extra-index-url https://pypi.fury.io/arrow-nightlies --pre pyarrow`). The two tests in this file go about testing differently: One uses parameterize but I wasn't able to find a good way to pass in a Table _and_ while also guarding the test with `pytest.importorskip`. The second just runs through queries one-by-one in a more manual fashion. 

Feel free to chop this up as you see fit.

### DIFF
--- a/ibis_substrait/tests/integration/test_pyarrow.py
+++ b/ibis_substrait/tests/integration/test_pyarrow.py
@@ -1,0 +1,101 @@
+import ibis
+import pytest
+
+from ibis_substrait.compiler.core import SubstraitCompiler
+
+compiler = SubstraitCompiler()
+
+
+def get_table_provider(tbl):
+    """Create a table_provider that always returns tbl"""
+
+    def table_provider(names, schema):
+        return tbl
+
+    return table_provider
+
+
+def to_ibis_table(arrow_table, table_name="t1"):
+    """Create ibis Table from pyarrow Table"""
+    return ibis.table(
+        zip(arrow_table.schema.names, [str(type) for type in arrow_table.schema.types]),
+        name=table_name,
+    )
+
+
+def run_query(expression, tbl, pa, substrait):
+    compiled = compiler.compile(expression)
+    query_bytes = compiled.SerializeToString()
+    result = substrait.run_query(
+        pa.py_buffer(query_bytes), table_provider=get_table_provider(tbl)
+    )
+
+    results = result.read_all()
+    assert type(results) == pa.lib.Table
+
+    return results
+
+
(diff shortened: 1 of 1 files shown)


ibis_substrait/tests/integration/test_pyarrow.py
```diff
@@ -0,0 +1,101 @@
...
+            "c": ["x", "y", "z"],
+            "d": [True, False, True],
+        },
+    )
+]
+
+
+@pytest.mark.parametrize("input,output", testdata)
+def test_pyarrow_produces_correct_result(input, output):
+    pa = pytest.importorskip("pyarrow")
```
gforsyth, line 60: Since all the tests in here are going to be w.r.t. `pyarrow`, we can put the `importorskip`s at the top-level of the file and pytest will skip the entire file.

ibis_substrait/tests/integration/test_pyarrow.py
```diff
@@ -0,0 +1,101 @@
...
+        pa.py_buffer(query_bytes), table_provider=get_table_provider(tbl)
+    )
+
+    results = result.read_all()
+    assert type(results) == pa.lib.Table
+
+    return results
+
+
+# WIP: I'm not sure how best to parameterize the query too without a ref to the Table
```
gforsyth, line 39: Yeah, so I think if we do the top-level importorskip we can have the `output` be the expected pyarrow objects without worrying about import headaches

gforsyth (changes requested):
This is great, @amoeba !  I left a few comments that I think will help with the parameterization question.

gforsyth:
Hey @amoeba -- I'm rolling this in to a separate PR because it already caught one bug I hadn't squashed

ibis_substrait/tests/integration/test_pyarrow.py
amoeba: oh cool, for some reason I didn't think that'd work

ibis_substrait/tests/integration/test_pyarrow.py
amoeba: ah yep. that's a lot easier

amoeba:
Sounds good @gforsyth, thanks. I'll let you close this when you're ready.

gforsyth:
Merged in as part of #538 -- thanks @amoeba !

amoeba:
Thanks for the help @gforsyth.

PR #538, opened 2023-03-13
feat(udf): add initial support for UDFs

gforsyth:
What does this do?
Adds a dispatch rule for the `ElementWiseVectorizedUDF` op.

This dispatch does the following:
- Uses the ibis type annotations from the UDF decorator to set the expected input and output types of the scalar function.
- sets the URI for this extension (since it is an extension) to the value of the `udf_uri` attribute of the `SubstraitCompiler`.  This should be set at instantiation time.  If there is no available `udf_uri` on the compiler it will error out.
- Returns the substrait scalar function call 

This assumes that you have some other means of registering the UDFs with your execution engine -- there's an example of what that might look like in `ibis_substrait/tests/integrations/tests/test_pyarrow.py`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,28 @@ jobs:
       - run: poetry install --sync
       - run: poetry run pip install 'ibis-framework==4.1'
       - run: poetry run pytest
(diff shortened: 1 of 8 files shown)


jcrist (approved):
:shipit:

PR #554, opened 2023-03-23
fix: Fix emit output of projection node to be based off input table i…

icexelloss:
…nstead of source tables

I found an issue with projection emit outputting when doing a "asof_join -> filter -> projection"

My code looks like
```
t = t1.asof_join(t2, ...) 
t = t[t[col] > 0]
t = t.muate(y=t[x] * 2)
```

Without the patch the plan looks like
```
relations {
  root {
    input {
      project {
        common {
          emit {
            output_mapping: 8
            output_mapping: 9
            output_mapping: 10
            output_mapping: 11
            output_mapping: 12
            output_mapping: 13
            output_mapping: 14
          }
        }
        input {
          filter {
            input {
              filter {
                input {
                  filter {
                    input {
                      extension_multi {
                        common {
                          emit {
                            output_mapping: 0
                            output_mapping: 1
                            output_mapping: 2
                            output_mapping: 3
                            output_mapping: 6
                            output_mapping: 7
                          }
                        }
...
```
This is problematic because the output of `asof_join` has 6 columns, not 8

This is I think how the plan should look like:
```
relations {
  root {
    input {
      project {
        common {
          emit {
            output_mapping: 6
            output_mapping: 7
            output_mapping: 8
            output_mapping: 9
            output_mapping: 10
            output_mapping: 11
            output_mapping: 12
          }
        }
        input {
          filter {
            input {
              filter {
                input {
                  filter {
                    input {
                      extension_multi {
                        common {
                          emit {
                            output_mapping: 0
                            output_mapping: 1
                            output_mapping: 2
                            output_mapping: 3
                            output_mapping: 6
                            output_mapping: 7
                          }
                        }
```

Edit:
I think the issue is not only related to asof join - even with aggregation I think it might be problematic, e.g

source [0, 1, 2, 3, 4, 5]
-> aggregation (let's say group by a column with two metrics) [0, 1, 2]
-> filter
-> projection

when computing the emit output from the last projection, with the logic before the PR, since the parent rel of projection is not aggregation or does it have an emit output, it will use the "else" logic which compute from the source, and the emit output for the last projection would start from 6, where it should start from 3. Basically as long as there is any rel between the projection and source that would reduce the number of columns this would not work I believe.

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -822,6 +822,12 @@ def selection(
             elif output_mapping := rel.common.emit.output_mapping:
                 mapping_counter = itertools.count(len(output_mapping))
                 break
+            elif not isinstance(op.table.op(), ops.Join):
+                # If child table is join, we cannot reliably call schema due to
+                # potentially duplicate column names, so fallback to the orignal
+                # logic.
+                mapping_counter = itertools.count(len(op.table.op().schema))
+                break
         else:
             source_tables = _find_parent_tables(op)
             mapping_counter = itertools.count(sum(len(t.schema) for t in source_tables))


icexelloss:
Will update PR description see - want to see how test goes

icexelloss:
@jcrist @gforsyth 

This bug is time sensitive w.r.t our Q1 release timeline - can we chat about quick?

Currently my patch fails some tpch test. I am trying to debug that but I haven't set up ibis-substrait dev environment.

ibis_substrait/compiler/translate.py
```diff
@@ -822,6 +822,9 @@ def selection(
             elif output_mapping := rel.common.emit.output_mapping:
                 mapping_counter = itertools.count(len(output_mapping))
                 break
+            elif relname in ("project", "filter"):
+                mapping_counter = itertools.count(len(op.table.op().schema))
+                break
         else:
             source_tables = _find_parent_tables(op)
```
icexelloss, line 829: Is there a reason that we need to look at source tables instead of the output schema of the immediate child? (i.e., op.table)

icexelloss:
The error appears to be caused by an ibis table with duplicate column names after the tpch join - is there a way to work around it?

gforsyth:
@icexelloss -- do you have an example of what this is fixing? It isn't clear to me what the bug is

gforsyth:
> @icexelloss -- do you have an example of what this is fixing? It isn't clear to me what the bug is

sorry, I was on email and didn't see the edited PR description

gforsyth:
@icexelloss -- if your current patch is fixing the `asof_join` issue, we can merge this in if you can squash the commits and make `commitlint` happy

icexelloss:
@gforsyth Thanks let me clean this up and will let you know.

Btw, I think the issue is not only related to asof join - even with aggregation I think it might be problematic, e.g

source [0, 1, 2, 3, 4, 5]
-> aggregation (let's say group by a column with two metrics) [0, 1, 2]
-> filter 
-> projection

when computing the emit output from the last projection, with the logic before the PR, since the parent rel of projection is not aggregation or does it have an emit output, it will use the "else" logic which compute from the source, and the emit output for the last projection would start from 6, where it should start from 3.

gforsyth:
> when computing the emit output from the last projection, with the logic before the PR, since the parent rel of projection is not aggregation or does it have an emit output, it will use the "else" logic which compute from the source, and the emit output for the last projection would start from 6, where it should start from 3. Basically as long as there is any rel between the projection and source that would reduce the number of columns this would not work I believe.

yes, I think you're right.  I'll add a test for that to go along with this fix.

icexelloss:
@gforsyth This should be good to go - can you take another look please?

gforsyth (approved):
Thanks @icexelloss !  I'll put in a follow-up to add a few tests, but I'm confirmed that this works locally.

icexelloss:
> Thanks @icexelloss ! I'll put in a follow-up to add a few tests, but I'm confirmed that this works locally.

Tyvm!

PR #559, opened 2023-03-27
feat: add support for ibis 5.x

gforsyth:
My apologies in advance for how enormous this PR has gotten.
I will note that several thousand lines of the diff are the new json snapshot outputs and can be safely ignored.

It has gotten a bit too hairy trying to support 3 major versions of Ibis, so I've broken things up.

There's now `translator/base.py` which has most of the original implementation.
All of the dispatchers have been moved over to `singlemethoddispatch` and then the specific bits for Ibis 3, 4, and 5 are handled in child class implementations with selective overrides.
Because the class dispatchers are descriptors, this involves a pretty gross hack to inherit the parent class translation registry -- happy to get some feedback on ways to make it less gross.

On a positive (?) note, since these are now classes, we don't have to thread the `compiler` through everywhere, we can attach it to the class and use it when needed.
I've also removed the default passing of `expr` everywhere -- there's only one dispatch that handles expressions and all it does is call `.op()` and then pass it back in.

I've tried to balance not duplicating too much code against not being too clever -- so there are a few `isinstance` checks which allow for both `ir.Table` and `ops.TableNode`, for example, because everything else in that translation rule is the same.

These can be simplified when we drop support for Ibis 3 (at some point in the future).


gforsyth:
I'm going to break this down into smaller chunks because this is intractable.

PR #575, opened 2023-04-07
feat(version): add version to produced substrait plans

gforsyth:
We should've been doing this, but better late than never.

This automatically updates the hash of the current substrait rev in the `__init__.py` -- I figure we can add a test to make sure that rev corresponds to the version listed in that same file.

Tests are currently failing because the validator is unhappy -- I'm not sure if it's aware of the Version field at all, so it may be time to jettison it.


gforsyth:
I'll probably wait to merge this until after #559 is in

PR #578, opened 2023-04-14
chore(poetry): make sqlalchemy a test dependency only

gforsyth:
We aren't using it explicitly, so it shouldn't need to be included for
end-users.  We DO need it as a `test` dep so that we can filter/suppress
upstream warnings in the test suite.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -373,7 +373,7 @@ pyflakes = ">=3.0.0,<3.1.0"
 name = "greenlet"
 version = "2.0.1"
 description = "Lightweight in-process concurrent programming"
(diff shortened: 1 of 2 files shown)


gforsyth:
> It does currently depend on it, although I don't know that the code that it hits is actually used for anything.

nevermind, I forgot what repo I was in

gforsyth:
Huh.  So if you want to silence a warning from an upstream dependency you have to install said dependency? Or maybe that's just on `nix`...

gforsyth:
Merging this -- remaining status checks are defunct.

PR #579, opened 2023-04-17
test: add pytest-snapshot tests for TPC-H outputs

gforsyth:


ibis_substrait/tests/compiler/test_tpch.py
```diff
@@ -704,8 +705,12 @@ def test_tpch1(tpc_h01, lineitem, compiler):
     "query",
     TPC_H,
 )
-def test_compile(query, compiler):
-    _ = compiler.compile(query)
+def test_compile(query, compiler, snapshot, request):
+    plan = compiler.compile(query)
+
+    snapshot.assert_match(
+        json_format.MessageToJson(plan), f"{request.node.callspec.id}.json"
```
jcrist, line 712: Is the output of this guaranteed to be consistent (key ordering, ...)? It looks like there's a few [options we could set](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.MessageToJson) (`sort_keys=True` stands out), but if the output order is already consistent then that may not matter.

jcrist (approved):
Just a question, otherwise LGTM.

ibis_substrait/tests/compiler/test_tpch.py
gforsyth: I don't know if it's guaranteed or not, but as I've been working across different versions of substrait and Ibis, the field ordering matches the ordering in the proto definitions (as least for the defined fields).

I just ran it with `sort_keys=True` and it's hard to read, actually, after having spent long enough looking at the substrait plans in their "normal" form

gforsyth:
Merging in b/c/ the branch protection rules are out-of-date (and were a tad premature)

PR #580, opened 2023-04-17
feat(version): add version to produced substrait plans

gforsyth:
We should've been doing this, but better late than never.

This automatically updates the hash of the current substrait rev in the `__init__.py` -- I figure we can add a test to make sure that rev corresponds to the version listed in that same file.

The substrait validator isn't aware of the version field at all and so can no longer parse these plans -- unfortunately there doesn't seem to be a way to make it ignore the error (since there is no associated error code)


gforsyth:
@cpcloud -- would you mind taking a quick look at this and confirming that I haven't built a rube goldberg machine to get the hash from nix?

cpcloud (approved):
LGTM! Not too bad!

gforsyth:
:pray: thanks for the check!

PR #583, opened 2023-04-18
feat: add support for std_dev and variance

### LINKED ISSUE #581: feat: add mappings to `std_dev` and `variance`

ianmcook:
Resolves #581

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -649,7 +649,7 @@ def _variance_base(
         value=translate(op.arg.op(), compiler=compiler, **kwargs)
     )
     translated_how = stalg.FunctionOption(
-        name='distribution',
+        name="distribution",
         preference=["POPULATION" if op.how == "pop" else "SAMPLE"]
     )
     return stalg.AggregateFunction(
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -645,14 +645,17 @@ def _variance_base(
 ) -> stalg.AggregateFunction:
     if compiler is None:
         raise ValueError
-    if op.how == "pop":
-        raise NotImplementedError("`how='pop'` not yet implemented")
     translated_arg = stalg.FunctionArgument(
         value=translate(op.arg.op(), compiler=compiler, **kwargs)
+    )
+    translated_how = stalg.FunctionOption(
+        name='distribution',
+        preference=["POPULATION" if op.how == "pop" else "SAMPLE"]
     )
     return stalg.AggregateFunction(
         function_reference=compiler.function_id(op=op),
         arguments=[translated_arg],
+        options=[translated_how],
         sorts=[],  # TODO: ibis doesn't support this yet
         phase=stalg.AggregationPhase.AGGREGATION_PHASE_INITIAL_TO_RESULT,
         output_type=translate(op.output_dtype),
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -649,8 +649,7 @@ def _variance_base(
         value=translate(op.arg.op(), compiler=compiler, **kwargs)
     )
     translated_how = stalg.FunctionOption(
-        name="distribution",
-        preference=["POPULATION" if op.how == "pop" else "SAMPLE"]
+        name="distribution", preference=["POPULATION" if op.how == "pop" else "SAMPLE"]
     )
     return stalg.AggregateFunction(
         function_reference=compiler.function_id(op=op),
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -634,6 +634,31 @@ def _count(
     )
 
 
+@translate.register(ops.StandardDev)
+@translate.register(ops.Variance)
+def _variance_base(
+    op: ops.StandardDev | ops.Variance,
+    expr: ir.ValueExpr | None = None,
+    *,
+    compiler: SubstraitCompiler | None = None,
+    **kwargs: Any,
+) -> stalg.AggregateFunction:
+    if compiler is None:
+        raise ValueError
+    if op.how == "pop":
+        raise NotImplementedError("`how='pop'` not yet implemented")
+    translated_arg = stalg.FunctionArgument(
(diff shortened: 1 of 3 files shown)


ianmcook:
@gforsyth When I run
```py
import ibis
from ibis_substrait.compiler.core import SubstraitCompiler

t = ibis.table(
    [("a", "string"), ("b", "float"), ("c", "int32"), ("d", "int64"), ("e", "int64")],
    "t",
)

expr = t.group_by(["a", "b"]).aggregate([t.c.var().name("var")]).select("b", "var")

compiler = SubstraitCompiler()
proto = compiler.compile(expr)
```
I see
```
Traceback (most recent call last):
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/core.py", line 83, in function_id
    function_extension = self.function_extensions[op_name]
KeyError: 'variance'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/core.py", line 197, in compile
    input=translate(expr.op(), expr=expr, compiler=self, **kwargs),
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/functools.py", line 889, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/translate.py", line 1080, in aggregation
    measures=[
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/translate.py", line 1082, in <listcomp>
    measure=translate(metric, compiler=compiler, **kwargs)
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/functools.py", line 889, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/translate.py", line 518, in alias_op
    return translate(op.arg.op(), alias_op.to_expr(), compiler=compiler, **kwargs)
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/functools.py", line 889, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/translate.py", line 600, in _reduction
    function_reference=compiler.function_id(expr=expr),
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/core.py", line 87, in function_id
    ] = self.extension_lookup(op_name, op)
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/core.py", line 113, in extension_lookup
    [
  File "/Users/ian/ibis-substrait/ibis_substrait/compiler/core.py", line 114, in <listcomp>
    IBIS_SUBSTRAIT_TYPE_MAPPING[arg.op().output_dtype.name]
AttributeError: 'str' object has no attribute 'op'
```
What step did I miss?

gforsyth:
Ahh, yeah, so `Variance` at least isn't the super-easy kind of op to add, but it should be relatively straightforward.  And `StandardDev` should be similar in effort.

If you look at the Ibis source for those ops, you can see that they both inherit from `VarianceBase`

https://github.com/ibis-project/ibis/blob/master/ibis/expr/operations/reductions.py#L144-L164

and while the first argument is any numeric arg type, the second is the sampling method (the `how` arg).

so you'll want to add a dispatch rule that looks something like


```
# you can decorate the same dispatch because the arguments are the same
@ops.StandardDev
@ops.Variance
def _variance_base(
    op: ops.StandardDev | ops.Variance,
    expr: ir.ValueExpr | None = None,
    *,
    compiler: SubstraitCompiler | None = None,
    **kwargs: Any,
) -> stalg.Expression:
    op = _check_and_upcast(op)
    # given the details of `op` -> function id
    return stalg.Expression(
        scalar_function=stalg.Expression.ScalarFunction(
            function_reference=compiler.function_id(op=op),
            output_type=translate(op.output_dtype),
            arguments=[
                stalg.FunctionArgument(
                    value=translate(arg, compiler=compiler, **kwargs)
                )
                for arg in op.args  # <-- need to make this not `op.args`
                if isinstance(arg, (ir.Expr, ops.Value))
            ],
        )
    )

```

and then construct and return a substrait expression (`stalg.Expression.ScalarFunction`) that only passes in the `op.arg` to the `argument` field.
And if `op.how` is something besides `'sample'` then you can either raise `NotImplementedError`, or we can go through passing on optional arguments to a scalar function. (I don't think we've had to do that thus far).

I'll think about how to make this easier because right now it does mean custom rules for all of these functions, but ibis holds "top level" and optional arguments in `op.args` and I don't know that there's a simple way to split those.

gforsyth:
If anyone comes across this PR, note that my example above is incorrect as `std` and `variance` are aggregate functions, not scalar functions, but the general idea still applies.

ibis_substrait/compiler/translate.py
```diff
@@ -634,6 +634,31 @@ def _count(
...
+        raise ValueError
+    if op.how == "pop":
+        raise NotImplementedError("`how='pop'` not yet implemented")
+    translated_arg = stalg.FunctionArgument(
+        value=translate(op.arg.op(), compiler=compiler, **kwargs)
+    )
+    return stalg.AggregateFunction(
+        function_reference=compiler.function_id(op=op),
+        arguments=[translated_arg],
+        sorts=[],  # TODO: ibis doesn't support this yet
```
ianmcook, line 656: @gforsyth If I wanted to hard-code something here to set the `distribution` option to `SAMPLE` in the plan, how would I do that?
```py
options = ???
```

icexelloss: I think Sth like
```
options = stalg.FunctionOption(name='distribution', preference=['SAMPLE'])
```

gforsyth: On mobile but I believe Li is correct.

Note that since 'sample' is the first option in the yaml files it will be chosen by default if nothing is set (or so says substrait convention)

ianmcook: Thanks, yes that worked. I implemented support for `how="pop"` in 589c882.

gforsyth (approved):
This looks good, thanks @ianmcook !

Issue #595, opened 2023-05-10
feat: add first/last aggregators


rtpsw:
Following the recent addition of [first/last reductions in Ibis](https://github.com/ibis-project/ibis/commit/8c01980ed5a30912ea35fa47903347ebf44c8b12), this issue will add first/last aggregators.

icexelloss:
This can be closed

gforsyth:
Closed via #600

PR #596, opened 2023-05-10
GH-595: feat: add first/last aggregators

rtpsw:
See https://github.com/ibis-project/ibis-substrait/issues/595


rtpsw:
cc @icexelloss

icexelloss:
Let's hold off on this. Per discussion with Weston we don't have to change [ibis_substrait/extensions/functions_aggregate_generic.yaml here
https://github.com/apache/arrow/pull/35513#issuecomment-1544181944

gforsyth:
Hey @rtpsw and @icexelloss -- just catching up after vacation.

The method here of adding entries to the ops mapping is correct, but the extension yaml files are vendored in from upstream.  As per some of the discussion in the linked Arrow issue, we'll want to either add those aggregations to the upstream extension definitions or we can use this to test out custom extensions.

rtpsw:
Yes, the yaml declarations for these aggregators will eventually be placed upstream. While this is pending, we tried to move forward with registering them as extensions for use in Arrow, which requires using the prefix `urn:arrow:substrait_simple_extension_function` for registration. We tried to use `[register_extension_yaml](https://github.com/ibis-project/ibis-substrait/blob/a48fa26251959545ed8123b2bc770186ad8faaaf/ibis_substrait/compiler/mapping.py#L183)` but there is a problem with [this line](https://github.com/ibis-project/ibis-substrait/blob/a48fa26251959545ed8123b2bc770186ad8faaaf/ibis_substrait/compiler/mapping.py#L198), which sets up `func.uri`. When we pass in a path to a `file.yaml`, containing the declarations for these extensions, and the above prefix then `func.uri` is set incorrectly. For `first`, it is set to `urn:arrow:substrait_simple_extension_function/file.yaml#first` but we Arrow needs it to be set to `urn:arrow:substrait_simple_extension_function#first` (as we verified).

rtpsw:
The discussion about extension functions has moved to #597.

icexelloss:
We can close this

gforsyth:
Closing this out.  We can revisit when the extensions land upstream.

PR #600, opened 2023-05-16
fix(uri): allow overriding extension uri when registering extension file

### LINKED ISSUE #597: feat: Support extension function outside of substrait spec
Substrait function is defined by the function uri + function name. 

Currently, all functions in ibis-strait comes from https://github.com/substrait-io/substrait/tree/main/extensions

There is a need to add function that of outside of main substrait spec. e.g., during https://github.com/apache/arrow/pull/35513#issuecomment-1544181944 to add first and last aggregation in Acero, we recognized that there is a need to communicate functions with alternatively uri, e.g., `urn:arrow:substrait_simple_extension_function` that is Acero specific. Other than the "first and last" function, we are also thinking about adding "floor and ceil" function for timestamp columns (e.g., in pandas, this is `s.dt.ceil` and `s.dt.floor`).

This ask here is that if we can allow an mechanism to support this. Currently, we can get this working by implementing some mechanism in our downstream codebase, and effectively registering first/last doing sth similar to this (https://github.com/ibis-project/ibis-substrait/blob/main/ibis_substrait/compiler/mapping.py#L183), but apparently that uses private functions in ibis-substrait and not a great solution.

@rtpsw Can share more details on what is done internally and we can discuss if we can have a better way to do this in ibis-substrait.

gforsyth:
cc @icexelloss @rtpsw will this do the trick?

Fixes #597

### DIFF
--- a/ibis_substrait/compiler/mapping.py
+++ b/ibis_substrait/compiler/mapping.py
@@ -180,8 +180,23 @@ def _parse_func(entry: Mapping[str, Any]) -> Iterator[FunctionEntry]:
         yield sf
 
 
-def register_extension_yaml(fname: str | Path, prefix: str | None = None) -> None:
-    """Add a substrait extension YAML file to the ibis substrait compiler."""
+def register_extension_yaml(
+    fname: str | Path, prefix: str | None = None, uri: str | None = None
+) -> None:
+    """Add a substrait extension YAML file to the ibis substrait compiler.
+
+    Parameters
+    ----------
+    fname
+        The filename of the extension yaml to register.
+    prefix
+        Custom prefix to use when constructing Substrait extension URI
+    uri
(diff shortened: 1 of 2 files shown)


icexelloss:
I think so. @rtpsw Can you check if this works for our purpose?

rtpsw:
@gforsyth, this works. Could this feature be quickly covered in a test?

icexelloss:
@gforsyth If you don't mind. It'd be great if we can merge this first and add tests in a follow up PR if needed so we can pull this change in Today.

gforsyth:
> @gforsyth If you don't mind. It'd be great if we can merge this first and add tests in a follow up PR if needed so we can pull this change in Today.

no worries, I just pushed up a test and I'm merging on green.

icexelloss:
Thanks!

PR #656, opened 2023-06-20
feat(substrait-python): use upstream substrait python package

gforsyth:
This is a breaking change.

We now rely on the upstream `substrait`
package (https://github.com/substrait-io/substrait-python) for the
generated substrait code.

All tests are passing -- this also accompanies a bump of the Substrait
version to v0.30.0 (because this is what is available in the upstream).


gforsyth:
merging into a release branch to avoid cutting two successive major releases (because we're also about to drop python 3.8)

PR #675, opened 2023-06-28
feat(substrait-python): use upstream substrait python package

gforsyth:
BREAKING CHANGE: We now rely on the upstream `substrait`
package (https://github.com/substrait-io/substrait-python) for the
generated substrait code. You will need to install `substrait` from pypi
or conda-forge if you have not already done so.

All tests are passing -- this also accompanies a bump of the Substrait
version to v0.30.0 (because this is what is available in the upstream).

Also dropping Python 3.8 support

* [x] Add `python-substrait` conda-forge package


gforsyth:
This was handled in #656

PR #690, opened 2023-07-05
ci: switch to manual release trigger

gforsyth:
I'm piling in too much stuff in #675 to avoid multiple major version bumps with the current automatic release process.

Rather than create a horrible PR over there, I'm going to switch us over to a push-button release process here.


jcrist (approved):

cpcloud (approved):
Nice!

PR #692, opened 2023-07-05
refactor: drop support for Ibis 3.x 

gforsyth:
BREAKING CHANGE: The minimum supported version of Ibis is 4.0.
Deprecated methods `sort_by` and `groupby` have been removed in Ibis
6.x, so it is not possible to support 3.x through 6.x.

Depends on #690 #691 

Also removing the decompiler, which is no longer supported or maintained.

I'll rebase this once the above PRs are in and it will be much easier to review.


gforsyth:
Rebased and ready for review!

PR #694, opened 2023-07-06
feat: add support for Ibis 6.x and maintain support for ibis 3.x

gforsyth:
I think this is gross, but I thought it would be even grosser.  Just putting this up to see if CI will even tolerate it (it is very difficult to install ibis 3.x at this point)


gforsyth:
Nope not doing this

PR #779, opened 2023-10-04
fix: check for function impls before trying to parse

richtia:
The recent addition of https://substrait.io/extensions/functions_geometry/
is breaking the parsing because the first dictionary entry is a custom data type
rather than an actual function.  The data type has no impls.

```
    for impl in entry["impls"]:
E   KeyError: 'impls'
```


richtia:
@gforsyth

gforsyth:
Thanks @richtia!  Glad someone else noticed this!  I just merged in the fix in #775 so I'm going to close this out.

PR #853, opened 2023-12-28
feat: support ibis 7.2

rshkv:
This PR adds support for Ibis 7.1 and 7.2.

I was trying out `ibis-substrait` with the latest `ibis-framework` (7.2) and hit the errors below. There were some breaks from 7.0 to 7.1 and 7.2.

<details>
<summary>Errors</summary>

With Ibis 7.1:
```
File ~/Library/Caches/pypoetry/virtualenvs/views-python-9ffzjCRm-py3.11/lib/python3.11/site-packages/ibis_substrait/compiler/translate.py:1333

-> 1333 @translate.register(ops.NotExistsSubquery)
   1334 def _not_exists_subquery(
   1335     op: ops.NotExistsSubquery,

AttributeError: module 'ibis.expr.operations' has no attribute 'NotExistsSubquery'
```
With Ibis 7.2:
```
File ~/Library/Caches/pypoetry/virtualenvs/views-python-9ffzjCRm-py3.11/lib/python3.11/site-packages/ibis_substrait/compiler/translate.py:27
     25 import toolz
     26 from ibis import util
---> 27 from ibis.common.graph import toposort
     28 from packaging import version

ImportError: cannot import name 'toposort' from 'ibis.common.graph' (/Users/wraschkowski/Library/Caches/pypoetry/virtualenvs/views-python-9ffzjCRm-py3.11/lib/python3.11/site-packages/ibis/common/graph.py)
```
</details>

I tested my changes locally with 4.0, 7.0, 7.1, and 7.2.


rshkv:
Poke

ibis_substrait/compiler/translate.py
```diff
@@ -932,7 +933,7 @@ def _find_parent_tables(op: ops.Selection) -> set[ops.PhysicalTable]:
     # a schema"
     source_tables = {
         t
-        for t in toposort(op).keys()
+        for t in Graph(op).toposort().keys()  # type: ignore[no-untyped-call]
```
rshkv, line 936: The top-level `toposort` function got dropped [here](https://github.com/ibis-project/ibis/pull/7758/files#diff-8516547b54230f69120865c854f04ac1815899eda6a1df2da6e5174b79d9c53aL476). It was delegating to `Graph.toposort` under the hood.

rshkv: Mypy was complaining about the `Graph` constructor being untyped. Is there a better fix than suppressing?
```
ibis_substrait/compiler/translate.py: note: In function "_find_parent_tables":
ibis_substrait/compiler/translate.py:936: error: Call to untyped function "Graph" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 2 source files)
```

ibis_substrait/compiler/translate.py
```diff
@@ -1174,8 +1175,8 @@ def _contains(
 
 if IBIS_GTE_7:
     # Contains was decomposed into two separate ops in Ibis 7.x
-    translate.register(ops.InColumn)(_contains)
-    translate.register(ops.InValues)(_contains)
+    translate.register(ops.InColumn)(_contains)  # type: ignore[no-untyped-call]
+    translate.register(ops.InValues)(_contains)  # type: ignore[no-untyped-call]
```
rshkv, line 1179: mypy: `register` on the `singledispatch` thingy is untyped.

ibis_substrait/compiler/translate.py
```diff
@@ -1369,6 +1369,11 @@ def _not_exists_subquery(
     )
 
 
+if not IBIS_GTE_71:
+    # NotExistsSubquery was removed in Ibis 7.1
+    translate.register(ops.NotExistsSubquery)(_not_exists_subquery)  # type: ignore[no-untyped-call]
```
rshkv, line 1374: `NotExistsSubquery` was dropped [here](https://github.com/ibis-project/ibis/pull/7344).

rshkv: Can move the `_not_exists_subquery` definition under the `if not IBIS_GTE_71` condition but wanted to avoid the whitespace diff.

ibis_substrait/compiler/translate.py
```diff
@@ -1449,6 +1454,9 @@ def _check_and_upcast(op: ops.Node) -> ops.Node:
     """Check that arguments to extension functions have consistent types."""
     op_name = IBIS_SUBSTRAIT_OP_MAPPING[type(op).__name__]
 
+    if any(isinstance(arg, ops.TableArrayView) for arg in op.args):
+        raise NotImplementedError("Subqueries are unsupported as function inputs")
+
```
rshkv, line 1459: Would ask you to double check my thinking here.

Before 7.1, `tpc_h17` was failing because `_rewrite_filter` wasn't implemented for `Aggregation`. 

With 7.1 [(this change)](https://github.com/ibis-project/ibis/pull/7368/), we make it past the rewrite filter stuff but `tpc_h17` still fails [because of this assertion](https://github.com/ibis-project/ibis-substrait/blob/4e8f2eb034a99c865b2e57fb8f786c750e9e893f/ibis_substrait/compiler/translate.py#L811-L813). It seems that something isn't right with selections that rely on subqueries.

We get similar failures on other TPC query tests with subqueries (H2 and H15). And I gather from the `xfail` annotations that Substrait translation isn't entirely implemented for subqueries and continues to not work with 7.1+, i.e. it's not something we need to fix _for_ 7.1.

I added this assertion to explicitly "unsupport" subquery translation (instead of implicitly failing on [that assertion](https://github.com/ibis-project/ibis-substrait/blob/4e8f2eb034a99c865b2e57fb8f786c750e9e893f/ibis_substrait/compiler/translate.py#L811-L813) as we do for other queries).

This obviously has the risk of failing something that wasn't failing before. Happy to undo and `xfail` H17 on an `AssertionError` same as H2 and H15.

ibis_substrait/tests/compiler/snapshots/test_tpch/test_compile/tpc_h092/tpc_h092.json
```diff
@@ -1121,7 +1109,7 @@
     }
   ],
   "version": {
-    "minorNumber": 24,
+    "minorNumber": 35,
```
rshkv, line 1112: The larger Substrait version bump here is because `tpc_h09.json` was unused. Tests relied on `tpc_h090.json` and `tpc_h091.json` instead.

ibis_substrait/tests/compiler/snapshots/test_tpch/test_compile/tpc_h092/tpc_h092.json
```diff
@@ -924,25 +924,13 @@
...
-                                              },
-                                              "rootReference": {}
+                                        "selection": {
+                                          "directReference": {
+                                            "structField": {
+                                              "field": 26
                                             }
                                           },
-                                          "failureBehavior": "FAILURE_BEHAVIOR_THROW_EXCEPTION"
+                                          "rootReference": {}
```
rshkv, line 933: This is the only diff relative to `tpc_h090.json`:

Screenshot 2023-12-30 at 16 11 10

ibis_substrait/tests/compiler/snapshots/test_tpch/test_compile/tpc_h111/tpc_h111.json
```diff
@@ -0,0 +1,515 @@
...
+                                      }
+                                    },
+                                    {
+                                      "value": {
+                                        "cast": {
+                                          "type": {
+                                            "decimal": {
+                                              "scale": 2,
+                                              "precision": 15,
+                                              "nullability": "NULLABILITY_NULLABLE"
```
rshkv, line 389: This line is the only diff relative to `tpc_h110.json`: 

Screenshot 2023-12-30 at 16 13 06

This changed because join outputs are now nullable because of [this change](https://github.com/ibis-project/ibis/pull/7596/files#diff-63b3fef44a98e55148586e372709a1c87c52dc23d3053db8654847306927055bR288).

ibis_substrait/tests/compiler/test_tpch.py
```diff
@@ -646,10 +644,27 @@ def tpc_h22(customer, orders):
             reason="Aggregates need to be handled differently than they are",
         ),
     ),
-    pytest.param(lazy_fixture("tpc_h09"), marks=ibis5),
-    pytest.param(lazy_fixture("tpc_h09"), marks=ibis4),
+    pytest.param(
+        lazy_fixture("tpc_h09"),
+        marks=ibis_version("<7.0,~=7.1.0", "Field ordering difference"),
```
rshkv, line 649: Weirdly, H9 on Ibis `<7` has the _same_ field order as `7.1.*`. Then `7.0.*` and `7.2.*` have the same field order. But `7.2` casts ints before multiplying them with decimals.

ibis_substrait/tests/compiler/test_tpch.py
```diff
@@ -673,16 +688,36 @@ def tpc_h22(customer, orders):
...
+    ),
+    pytest.param(
+        lazy_fixture("tpc_h17"),
+        marks=(
+            pytest.mark.xfail(
+                raises=NotImplementedError,
+                reason="Filters on subqueries are unsupported",
+            ),
+            ibis_version(">=7.1", "Aggregation not implemented before 7.1"),
         ),
```
rshkv, line 707: See https://github.com/ibis-project/ibis-substrait/pull/853/files#r1438625638.

ibis_substrait/compiler/translate.py
gforsyth: I think that your analysis here is correct and this is a reasonable solution.  We do need to do a more thorough job with subqueries, but we're currently rewriting a bunch of Ibis internals (ongoing) so I'm good with leaving this in an explicitly unsupported state for now.

gforsyth (approved):
Hey @rshkv, thank you for putting this in!

Sorry for the delay in reviewing -- I was off over the holidays and I'm just getting back to this. 

I'm on board with all of these changes and we can hopefully tackle better subquery support soon!

PR #1078, opened 2024-07-22
fix(field_index): get field index w.r.t. pre-join table schemata

### LINKED ISSUE #1072: bug: Aggregation field index off by one when joining two tables.
### What happened?

With the following query
```
        expr = (
            tests.inner_join(vehicles, vehicles.vehicle_id == tests.vehicle_id)        
            .group_by(vehicles.colour)
            .count()
            .select("colour", "CountStar()")
        )
```

This produces a plan; but the field offset used for the aggregation is 'off-by-one':   For example here's the plan created (I've a custom formatted for substrait)

```
<Substrait Plan>
Root::  ImmutableProject [colour, CountStar()]
<Substrait Relation>
+- Project:: [Str, I64, Str, I64]
   +- Aggregate:: FieldRef#/Str/StructField{offset=10}
      +- Join:: INNER <ScalarFn>equal:any_any
          : arg0 = FieldRef#/I32/StructField{offset=8}
          : arg1 = FieldRef#/I32/StructField{offset=1}
         +- NamedScan::  Tables=[tests] Fields=test_id[I32],vehicle_id[I32],test_date[Str],test_class[Str],test_type[Str],test_result[Str],test_mileage[I32],postcode_area[Str]
         +- NamedScan::  Tables=[vehicles] Fields=vehicle_id[I32],make[Str],model[Str],colour[Str],fuel_type[Str],cylinder_capacity[I32],first_use_date[Str]
```

Note that the offset of the aggregate is 10, and counting up that's `model` not `colour`;  this only occurs if there's a join, and only with the second table. So aggregating on say `test_type` would be ok. So I'm wondering if the issue is that the `vehicle_id` which is the join key is perhaps being ignored?

Can also be seen just printing out the protobuf

```
            groupings {
              grouping_expressions {
                selection {
                  direct_reference {
                    struct_field {
                      field: 10
                    }
                  }
                  root_reference {
                  }
                }
              }
            }
```

Specification of the tables is 
```
        tests = ibis.table(
            [
                ("test_id", "int32"),
                ("vehicle_id", "int32"),
                ("test_date", "str"),
                ("test_class", "str"),
                ("test_type", "str"),
                ("test_result", "str"),
                ("test_mileage", "int32"),
                ("postcode_area", "str"),
            ],
            name="tests",
        )

        vehicles = ibis.table(
            [
                ("vehicle_id", "int32"),
                ("make", "str"),
                ("model", "str"),
                ("colour", "str"),
                ("fuel_type", "str"),
                ("cylinder_capacity", "int32"),
                ("first_use_date", "str"),
            ],
            name="vehicles",
        )
```

I've looked at source code; but not spotted how this index is calculated for the aggregation case. 

### What version of ibis-substrait are you using?

v4

### What substrait consumer(s) are you using, if any?

Quite a few of them!  in this case I saw it via Spark; but the problem can been seen without a consumer

### Relevant log output

_No response_

gforsyth:
JoinChains provide the schema of the joined table (which is great for Ibis)
but for substrait we need the Field index computed with respect to
the original table schemata.  In practice, this means rolling through
the tables in a JoinChain and computing the field index _without_
removing the join key

Given
```
Table 1
  a: int
  b: int

Table 2
  a: int
  c: int

JoinChain[r0]
 JoinLink[inner, r1]
   r0.a == r1.a
 values:
   a: r0.a
   b: r0.b
   c: r1.c
```

If we ask for the field index of `c`, the JoinChain schema will give
us an index of `2`, but it should be `3` because

```
 0: table 1 a
 1: table 1 b
 2: table 2 a
 3: table 2 c
```

So now we pull out the correct JoinReference object and use that to
index into the tables in the JoinChain and offset by the length of the
schema of those preceding tables.

Resolves #1072

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -712,19 +712,17 @@ def table_column(
         # Join reference containing the field we care about
         field_table = op.rel.values.get(op.name).rel
         # Index of that join reference in the list of join references
-        field_table_index = op.rel.tables.index(field_table)
+        field_table_index = join_tables.index(field_table)
 
-        join_table_offset = 0
         # Offset by the number of columns in each preceding table
-        for i in range(field_table_index):
-            join_table_offset += len(join_tables[i].schema)
+        join_table_offset = sum(
+            len(join_tables[i].schema) for i in range(field_table_index)
+        )
         # Then add on the index of the column in the table
         # Also in the event of renaming due to join collisions, resolve
         # the renamed column to the original name so we can pull it off the parent table
         orig_name = op.rel.values[op.name].name
-        relative_offset = join_table_offset + field_table.schema._name_locs.get(
-            orig_name
-        )
+        relative_offset = join_table_offset + field_table.schema._name_locs[orig_name]
     else:
         schema = op.rel.schema
         relative_offset = schema._name_locs[op.name]
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -674,8 +674,60 @@ def table_column(
     else:
         base_offset = 0
 
-    schema = op.rel.schema
-    relative_offset = schema._name_locs[op.name]
+    if isinstance(op.rel, ops.JoinChain):
+        # JoinChains provide the schema of the joined table (which is great for Ibis)
+        # but for substrait we need the Field index computed with respect to
+        # the original table schemas.  In practice, this means rolling through
+        # the tables in a JoinChain and computing the field index _without_
+        # removing the join key
+        #
+        # Given
+        # Table 1
+        #   a: int
+        #   b: int
+        #
+        # Table 2
+        #   a: int
+        #   c: int
+        #
+        # JoinChain[r0]
+        #  JoinLink[inner, r1]
+        #    r0.a == r1.a
+        #  values:
+        #    a: r0.a
+        #    b: r0.b
+        #    c: r1.c
+        #
+        # If we ask for the field index of `c`, the JoinChain schema will give
+        # us an index of `2`, but it should be `3` because
+        #
+        #  0: table 1 a
+        #  1: table 1 b
+        #  2: table 2 a
+        #  3: table 2 c
+        #
+
+        # List of join reference objects
+        join_tables = op.rel.tables
+        # Join reference containing the field we care about
+        field_table = op.rel.values.get(op.name).rel
+        # Index of that join reference in the list of join references
+        field_table_index = op.rel.tables.index(field_table)
+
+        join_table_offset = 0
+        # Offset by the number of columns in each preceding table
+        for i in range(field_table_index):
+            join_table_offset += len(join_tables[i].schema)
+        # Then add on the index of the column in the table
+        # Also in the event of renaming due to join collisions, resolve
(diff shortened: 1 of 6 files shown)


ibis_substrait/compiler/translate.py
```diff
@@ -674,8 +674,60 @@ def table_column(
...
+        #  2: table 2 a
+        #  3: table 2 c
+        #
+
+        # List of join reference objects
+        join_tables = op.rel.tables
+        # Join reference containing the field we care about
+        field_table = op.rel.values.get(op.name).rel
+        # Index of that join reference in the list of join references
+        field_table_index = op.rel.tables.index(field_table)
```
cpcloud, line 715:
```suggestion
        field_table_index = join_tables.index(field_table)
```

ibis_substrait/compiler/translate.py
```diff
@@ -674,8 +674,60 @@ def table_column(
...
+        # Offset by the number of columns in each preceding table
+        for i in range(field_table_index):
+            join_table_offset += len(join_tables[i].schema)
+        # Then add on the index of the column in the table
+        # Also in the event of renaming due to join collisions, resolve
+        # the renamed column to the original name so we can pull it off the parent table
+        orig_name = op.rel.values[op.name].name
+        relative_offset = join_table_offset + field_table.schema._name_locs.get(
+            orig_name
+        )
```
cpcloud, line 727:
```suggestion
        relative_offset = join_table_offset + field_table.schema._name_locs[orig_name]
```

Seems like this would fail with an inscrutable `NoneType` error if `orig_name` doesn't exist in `_name_locs`, whereas you'd actually want to know that the key isn't there.

ibis_substrait/compiler/translate.py
```diff
@@ -674,8 +674,60 @@ def table_column(
...
+        join_tables = op.rel.tables
+        # Join reference containing the field we care about
+        field_table = op.rel.values.get(op.name).rel
+        # Index of that join reference in the list of join references
+        field_table_index = op.rel.tables.index(field_table)
+
+        join_table_offset = 0
+        # Offset by the number of columns in each preceding table
+        for i in range(field_table_index):
+            join_table_offset += len(join_tables[i].schema)
```
cpcloud, line 720:
```suggestion
        # Offset by the number of columns in each preceding table
        join_table_offset = sum(len(join_tables[i].schema) for i in range(field_table_index))
```

Maybe it's just me, but I find the summation a bit easier to read.

ibis_substrait/tests/compiler/snapshots/test_tpch/test_compile/tpc_h07/tpc_h07.json
```diff
@@ -732,7 +732,7 @@
                                                       "selection": {
                                                         "directReference": {
                                                           "structField": {
-                                                            "field": 1
+                                                            "field": 45
```
cpcloud, line 735: A 45x improvement in accuracy!!

cpcloud (changes requested):
Looks good, only thing I think should definitely be changed is the `.get` call!

ibis_substrait/compiler/translate.py
gforsyth: Very good point.

PR #1108, opened 2024-08-22
feat: add datafusion to parity tests

tokoko:
- adds datafusion to parity tests
- minor changes due to a breaking join type change in substrait

### DIFF
--- a/ibis_substrait/tests/compiler/test_parity.py
+++ b/ibis_substrait/tests/compiler/test_parity.py
@@ -76,9 +76,38 @@ def run_query_duckdb_substrait(expr, datasets, compiler):
         return result.fetch_arrow_table()
 
 
+def run_query_datafusion(expr, datasets, compiler):
+    import pyarrow
+    import datafusion.substrait
+
+    connection = datafusion.SessionContext()
+
+    for k, v in datasets.items():  # noqa: B007
+        connection.deregister_table(k)
+        connection.register_record_batches(k, [v.to_batches()])
+
+    plan = compiler.compile(expr)
+
+    plan_data = plan.SerializeToString()
+    substrait_plan = datafusion.substrait.serde.deserialize_bytes(plan_data)
+    logical_plan = datafusion.substrait.consumer.from_substrait_plan(
+        connection, substrait_plan
+    )
+
+    df = connection.create_dataframe_from_logical_plan(logical_plan)
+    for column_number, column_name in enumerate(df.schema().names):
+        df = df.with_column_renamed(
+            column_name, plan.relations[0].root.names[column_number]
+        )
+    return df.to_arrow_table()
+
+
 def run_parity_tests(expr, datasets, compiler, engines=None):
     if engines is None:
-        engines = ["acero"]  # duckdb_substrait disabled because can't run on windows
+        engines = [
+            "acero",
+            "datafusion",
+        ]  # duckdb_substrait disabled because can't run on windows
     res_duckdb = sort_pyarrow_table(run_query_duckdb(expr, datasets))
     if "acero" in engines:
         res_acero = sort_pyarrow_table(run_query_acero(expr, datasets, compiler))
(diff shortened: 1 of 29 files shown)


gforsyth:
Thanks for putting this in @tokoko !  If you can rebase against `main`, should be good to go

tokoko:
@gforsyth should be all good now. thanks

ibis_substrait/tests/compiler/test_parity.py
```diff
@@ -176,7 +210,7 @@ def test_filter_groupby():
     )
 
     compiler = SubstraitCompiler()
-    run_parity_tests(grouped_table, datasets, compiler=compiler)
+    run_parity_tests(grouped_table, datasets, compiler=compiler, engines=["acero"])
```
gforsyth, line 213: Now that there are multiple engines running, we might want to be more deliberate in tracking what's working and what isn't.  I'm reading this line as implicitly saying that "datafusion doesn't pass this test" -- if that's the case, maybe we're better off parametrizing each of these tests by `engine` and then marking the individual failing cases?

tokoko: yes, that's right. I'll probably spin off engines as fixtures and parameterize + mark failing cases.

tokoko: Looks like I'll have to do some refactoring to do that, let's hold this for a little while until I do that. I'll spin off engines as separate components. btw, Ideally I was thinking of relying on work already done in [spark-substrait-gateway](https://github.com/voltrondata/spark-substrait-gateway/tree/main/src/backends). It's more polished and handles some engine-specific corner cases as well.

tokoko: Made the changes, had to get a little creative around markings. please take a look and let me know what you think.

ibis_substrait/tests/compiler/test_parity.py
```diff
@@ -117,127 +69,164 @@ def run_parity_tests(expr, datasets, compiler, engines=None):
...
+        },
+    ),
+    ("ibis_aggregate_having", {"datafusion_consumer": (Exception, "")}),
+    "ibis_inner_join_chain",
+    "ibis_union",
+    ("ibis_window", {"acero_consumer": (pa.ArrowNotImplementedError, "Unimplemented")}),
+    "ibis_is_in",
+    (
+        "ibis_scalar_subquery",
+        {"acero_consumer": (pa.ArrowNotImplementedError, "Unimplemented")},
```
gforsyth, line 201: I think this is probably better handled with `pytest.param` -- so it's likely that we'll need to slightly reorganize how these are being tested.

gforsyth (changes requested):
Thanks @tokoko !

I think that the list of fixture-names and custom parameter checking is probably going to be more of a maintenance burden than we want to deal with.

I think the things you are actually _testing_ are great, but we can probably rearrange them in a more manageable way.

I think all of the current Ibis expression `fixtures` should be the tests themselves -- then those can each be parametrized over consumers. 

The execution and comparison code seems sufficiently generic that we can pull that out into a little helper function, so you don't have to copy that everywhere.

tokoko:
Sure, It actually started out that way, but I tried to mimic tpch tests and reorganized lol. I'll try to split them up and see where that goes.

tokoko:
@gforsyth Made the changes, reorganized each query as a separate test function. please TAL when you get the chance.

gforsyth (commented):
Awesome.  Thanks for pushing on this @tokoko !

This layout looks much more readable to me.

I know that Datafusion has... suboptimal... exceptions -- are those test failures raising any other messages?

tokoko:
Cool.

> I know that Datafusion has... suboptimal... exceptions -- are those test failures raising any other messages?

Error messages are not that bad in this case, two of them are SchemaErrors iirc. I just used `Exception` to mean that I haven't yet made sense of the reason exactly why they are thrown yet. It's highly likely I'm doing something wrong in datafusion consumer. I was basically planning to come back to it later.

gforsyth (approved):

gforsyth:
> I was basically planning to come back to it later.

Sounds good to me.  Thanks again for working through this!

PR #1119, opened 2024-08-28
fix: handle projections after join chain

tokoko:
Ibis IR has a bit of a quirky behavior around projections, namely one can define projections both in `ops.Project` object and also as part of `ops.JoinChain`, not sure why that's the case. Current compiler handles only projections defined in `ops.Project`, therefore any selects written after a join chain are ignored.

The PR fixes the issue by injecting a projection node after every join chain in the plan. Technically this is not always necessary, because if the values contained in the `ops.JoinChain` do only column renames and reorders (no new expressions) then substrait's Emit message inside JoinRel can be used as well, but that would require inspecting all values first to make sure that nothing other than field references are present in there.

### DIFF
--- a/ibis_substrait/compiler/translate.py
+++ b/ibis_substrait/compiler/translate.py
@@ -818,11 +818,11 @@ def filter(
 def apply_projection(
     schema_len: int,
     relation: stalg.Rel,
-    values,
+    values: Mapping[str, ops.Value],
     compiler: SubstraitCompiler,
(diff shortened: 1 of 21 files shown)


gforsyth (approved):
Thanks for putting this in @tokoko !

Yeah, I suspect there are a few more changes in the Ibis IR that we aren't catching, but we can fix them as they arise.

PR #1129, opened 2024-09-05
fix: amend datafusion aggregations in parity tests

tokoko:
per #1120
- adds failing test for the bug in datafusion in which a plan with aggregate as the last node can't be executed.
- adds `.filter(ibis.literal(True)` after aggregations in other tests to work around a bug in datafusion.


gforsyth (approved):

gforsyth:
Thanks, @tokoko!

PR #1134, opened 2024-09-06
fix(extensions): prevent ambiguous scalar function lookups

gforsyth:
Truly a terrible piece of work.

The already slightly hacky extension loading / matching mechanism has failed because there are now key collisions because several yaml files have a function called `count` that takes `any` as the input type, but with _different_ return types.

I had also not been keeping up with the extension changes and we weren't handling a bunch of different extension type signatures.

This whole thing probably needs a rewrite, but I'm not going to do that right now.


gforsyth:
I'm merging this in to get things passing

tokoko:
> The already slightly hacky extension loading / matching mechanism has failed because there are now key collisions because several yaml files have a function called `count` that takes `any` as the input type, but with _different_ return types.

@gforsyth I think this was probably the wrong approach to take, namely including return type in the signature/lookup. According to the [docs](https://substrait.io/extensions/#function-signature-compound-names): "A YAML file may contain one or more functions by the same name. The key used in the function extension declaration to reference a function is a combination of the name of the function along with a list of the required input argument types"

imho what was missing to properly disambiguate between the functions was the file path, rather than a return type. I'm thinking of changing `_extension_mapping` to be a three-level dict (file name, function name, signature) and function mapping will have to be altered as a consequence to something like this (ideally more succinct than shown here):

```
IBIS_SUBSTRAIT_OP_MAPPING = {
    "Abs": ("https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml","abs"),
    "Acos": ("https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml", "acos"),
    ...
```

gforsyth:
Hey @tokoko -- I agree completely.  This was very much meant as a temporary measure, and I think your proposal is a good one.

tokoko:
Another point I wanted to quickly run by you. I'm also thinking of giving up on impl lookup approach once you narrow it down to a function. The current approach misses some valid impls, while doesn't check others strictly enough, namely the first [lookup](https://github.com/ibis-project/ibis-substrait/blob/main/ibis_substrait/compiler/core.py#L119) assumes that all args will be of type `any`, but actually there are some functions which have some args of type `any` and other of some concrete type (for example [lead](https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml#L1778)). On the other simple `any` lookups don't check the additional constraints that `any` types may have (`any1`, `any2` and so on).

In short, I'm thinking of instead adding a method to `FunctionEntry` class that will take in desired signature and return whether signature is valid for the entry or not. During a function "lookup", we will simply loop through all available impls of a function until at least one of them reports being valid.

gforsyth:
I think that sounds fine to me -- some of the current issues with assigning scalar function signatures stem from differences between `ibis` and `substrait` about what the output dtype of an function should be (just a question of _precision_).  I imagine that can be part of the validity checks for a given entry.

PR #1144, opened 2024-09-20
test: add tests for some arithmetic scalar functions

tokoko:
Adds tests for some arithmetic scalar functions. All passing functions are run as part of a single test for simplicity. Might make sense to split them up in the future if more rigorous testing is introduced (e.g. different data types)


gforsyth (approved):
:shipit: 

Thanks, @tokoko !

Issue #1145, opened 2024-09-20
feat: Annotate all Rels with output_names


tokoko:
substrait doesn't keep track of intermediate field names, which makes it hard to properly round-trip queries. Now there's a way to optionally include names in `output_names` field as a [hint](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L46) attached to every Rel. compiler should populate these fields in order for eventual decompiler to take advantage of them.

tokoko:
@gforsyth this will probably result in considerably larger proto objects for some queries, so I'm thinking of making the feature optional.

gforsyth:
Hey @tokoko!  I have a few open-ended questions that I think might be useful to flesh out a bit.

Would an eventual decompiler (oh no) _require_ these intermediate field names or merely take advantage of them when they were available?

If they're _required_, then I think making them optional is going to be a future foot-gun.

I guess this next one isn't a question: I wouldn't worry about making the feature optional until a) we've confirmed how much larger it makes proto objects (probably something like `O(num_rels)`?) and b) that we have some notion that there is actually a performance penalty worth caring about, like deserialization cost vs query execution time.

tokoko:
> Would an eventual decompiler (oh no) _require_ these intermediate field names or merely take advantage of them when they were available?

I think not. Decompiler should ideally handle all valid substrait plans, not just the ones that we produce, so it should be strictly optional with a fallback to "randomly" generated names.

I agree on the second point. We can make it work first and evaluate performance trade-off later.

PR #1150, opened 2024-09-24
feat: add outputNames hint to all Rels except joins

tokoko:
This is first PR to solve #1145 and adds outputNames hint to all Rels except for JoinRel and CrossRel. Join relations need more work to handle name collisions as ibis doesn't do it for every join step as part of JoinChain.


gforsyth (approved):
Thanks, @tokoko !

PR #1154, opened 2024-09-27
feat: add extension language interpreter

tokoko:
This is a sister PR to #1153, adding a simple interpreter for expression language needed for extension handling as described [here](https://substrait.io/expressions/scalar_functions/#return-type-expressions) (doesn't handle all grammar yet). The purpose is to then use this interpreter for return type definitions like [this](https://github.com/substrait-io/substrait/blob/1f3354d9f0f8b4425e98623f34d1e4578e2142bd/extensions/functions_arithmetic_decimal.yaml#L17).

- Adds a required dependency on [pyparsing](https://github.com/pyparsing/pyparsing) library, which is used to parse expressions. It's a no-dependency lightweigh library so I though the inclusion wouldn't be an issue.
- Just like `FunctionRegistry` this component is imho also better suited for `substrait-python` so implementation is intentionally ibis-free.

### DIFF
--- a/ibis_substrait/compiler/extension_language.py
+++ b/ibis_substrait/compiler/extension_language.py
@@ -1,0 +1,114 @@
+from typing import Optional, Union
+
+from pyparsing import (
+    Forward,
+    Group,
+    Literal,
+    ParseResults,
+    Word,
+    ZeroOrMore,
+    alphas,
+    identchars,
+    infix_notation,
+    nums,
+    oneOf,
+    opAssoc,
+    pyparsing_common,
+)
+
+supported_functions = ["max", "min"]
+
+multop = oneOf("* /")
+plusop = oneOf("+ -")
+compop = oneOf("> <")
+assign = Word(identchars + nums)("assign")
+variable = Word(identchars + nums)("variable")
+integer = Word(nums)
+expr = Forward()
+fn = Group(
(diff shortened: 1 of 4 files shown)


tokoko:
I was thinking of battle-testing if here first, but I guess that's probably a better idea. I asked in substrait slack and looks like there is antlr grammar in substrait-java which can be used instead of my pyparsing implementation in here. I'll go over the whole thing one more time to reflect that and open a PR in substrait-python.

gforsyth:
I mean, if the antlr grammar can be used easily, great, but since you already have the pyparsing implementation, that might be lighter-weight.  I guess we can see how it goes.

tokoko:
yeah, agreed. we'll have to see. pyparsing is a lot simpler with less moving parts, but antlr is language-agnostic and can turn into a central definition for different language implementations to use, which is probably a more important win.

gforsyth:
Maybe we can hide the java in the build process -- like generate the python-specific stubs and check those in as needed without requiring `substrait-java` as a dependency

tokoko:
we should definitely not depend on `substrait-java`, I'll just copy over antlr grammar file from them and toss in a bash script for python stubs generation. The only thing shared will be the grammar file once it's moved to the core substrait repo from substrait-java.

